PR #102, opened 2014-06-13
Fix bug with theme and lang using together

ExReanimator:
If using "recaptcha_tags display: { theme: 'white', lang: I18n.locale }" theme param not works and shown default (red).
With my fix it works good.


grosser:
smells like something else is broken ... I'll close this since it's 1 year old ...

PR #118, opened 2015-02-10
Support for language code param for the v2 API

clarif:
Support languages for the v2 API (the language param (hl) needs to be passed as a parameter when calling the JS API file).

Usage: `recaptcha_tags hl: 'fr'` (note: works only for the v2 API, e.g. `config.api_version = 'v2'`).


zakelfassi:
+1. Was about to add that.

PR #125, opened 2015-05-15
add style for v2 recaptcha_tags helper

finist:
add style attribute for g-recaptcha, for case like this https://www.geekgoddess.com/how-to-resize-the-google-nocaptcha-recaptcha/


grosser:
- use css ?
 - wrap it in a div that has styles ?

otherwise please rebase and add a test + docs in the readme

grosser:
FYI there is an example app now, so feel free to include a screenshot with how it would look :)

PR #126, opened 2015-05-28
Patch 1

jdrago999:


jdrago999:
After a while your session will expire.  The `expired-callback` argument provides the ability to respond to this event as described here: http://stackoverflow.com/a/29908562/4704953

ambethia:
Thanks!

PR #129, opened 2015-06-16
Switch to Rails.env in env_options

fenec:
RAILS_ENV is not always set in development mode, so it would be convenient to rely on Rails.env instead.

### DIFF
--- a/lib/recaptcha/verify.rb
+++ b/lib/recaptcha/verify.rb
@@ -6,7 +6,7 @@ module Recaptcha
     def verify_recaptcha(options = {})
       options = {:model => options} unless options.is_a? Hash
 
-      env_options = options[:env] || ENV['RAILS_ENV']
+      env_options = options[:env] || ENV['RAILS_ENV'] || Rails.env if defined? Rails.env
       return true if Recaptcha.configuration.skip_verify_env.include? env_options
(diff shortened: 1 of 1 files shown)


lib/recaptcha/verify.rb
```diff
@@ -6,7 +6,7 @@ module Verify
     def verify_recaptcha(options = {})
       options = {:model => options} unless options.is_a? Hash
 
-      env_options = options[:env] || ENV['RAILS_ENV']
+      env_options = options[:env] || ENV['RAILS_ENV'] || Rails.env if defined? Rails.env
```
marceloboeira: Do you need the `ENV['RAILS_ENV']`, `Rails.env` does not cover all?

grosser: damn ... this is actually wrong ... 

```
a = 1 || 2 if false
=> nil
```

grosser: having tests would be great :/

grosser: fixed by putting parens around it

grosser: Rails.env might be not defined ... so we need the ENV ... but we should be doing Rails.env || ENV

Issue #130, opened 2015-06-29
Version 2 from Arbitrary Domains


brendon:
Some of you may have been wondering how to use v2 of Recaptcha with an arbitrary domain. Apparently it's possible via a special attribute:

![developer guide for secure token 2](https://cloud.githubusercontent.com/assets/81871/8402826/069881ba-1e93-11e5-95cc-c2213a2763a6.jpg)

![developer guide for secure token 2 2](https://cloud.githubusercontent.com/assets/81871/8402839/1cdc59b0-1e93-11e5-84ae-bb8ee4843f34.jpg)

I was wondering if it would be possible to add this parameter as a runtime configuration option, or perhaps have the private key as a configuration parameter and have the rest calculated automatically for each request if it's present?

brendon:
Any chance of some movement on this?

ambethia:
Sorry, I completely hadn't looked at this yet. Can you add some test coverage for this?

brendon:
Hi @ambethia :) That's all good.

My question was basically, how would I test this? I had a hard time figuring our where it'd fit in your suite, and you seem to have a lot of those kind of tests skipped anyway?

The actual results shouldn't change (i.e. integration tests should still pass if the extra code works as intended), the only difference would be that you don't have to pre-specify domains on google's web interface beforehand (so this is hard to test automatically).

ambethia:
A lot of them are skipped, I'm just trying to stay ahead of not making that situation worse. Looking at the code, there really is no logic (that belongs to this library) to test, any tests would just be testing that the crypto stuff does it's job. As long as the existing test suite still passes, this shouldn't be a problem.

Thanks!

brendon:
Thanks Jason, glad to have that merged in! Would you be able to release a new gem? :)

Have a great day!

ambethia:
I need to look through some of the other open PRs then ill do a release, maybe this weekend?

brendon:
Sounds good to me :)

PR #132, opened 2015-07-01
Generate stoken and attach it as a data attribute

brendon:
With reference to https://github.com/ambethia/recaptcha/issues/130 this
change generates a secure_token based on a random uuid and the integer
time in milliseconds as a JSON message. This is encrypted using the
‘private_key’ and sent to google as a data attribute.


brendon:
I'm not sure where to start with the tests, but I was hoping someone could help me with that.

The code works and I've placed it into production using my fork.

PR #134, opened 2015-07-07
Update README with multiple widget example

bfcoder:
Adds multiple widget example as requested in https://github.com/ambethia/recaptcha/issues/79


ambethia:
Thanks!

PR #136, opened 2015-07-26
fix README config.api_version default value

vyachkonovalov:
According to <a href="https://github.com/ambethia/recaptcha/blob/master/lib/recaptcha.rb#L21" target="_blank">this</a>
```ruby
    RECAPTCHA_API_VERSION = 'v2' 
```
default value of config.api_version is 'v2', but README says the uncomment make sense.


grosser:
merged by hand, thanks!

PR #141, opened 2015-10-16
Add option to disable security token in API v2

facano:
Add option in recaptcha_tag to disable security token for API v2


grosser:
manually merged + made it not generate the stoken when it is not needed

PR #142, opened 2015-11-09
Break out secure token generation.

astuyve:
Basically just breaking out the logic to generate a secure token into it's own module.

I did this so that I can generate cross-domain recaptcha widgets without having to write my own secure token module.


lib/recaptcha/recaptcha_token.rb
```diff
@@ -0,0 +1,14 @@
+module RecaptchaToken
```
grosser: should be a namespace module

lib/recaptcha/recaptcha_token.rb
```diff
@@ -0,0 +1,14 @@
...
+  def self.secure_token
+    stoken_json = {'session_id' => SecureRandom.uuid, 'ts_ms' => (Time.now.to_f * 1000).to_i}.to_json
+    cipher = OpenSSL::Cipher::AES128.new(:ECB)
+    private_key_digest = Digest::SHA1.digest(Recaptcha.configuration.private_key)[0...16]
+
+    cipher.encrypt
+    cipher.key = private_key_digest
+    encrypted_stoken = cipher.update(stoken_json) << cipher.final
+    Base64.urlsafe_encode64(encrypted_stoken).gsub(/\=+\Z/, '')
+  end
```
grosser: since it's a new public api it needs at least some trivial tests

lib/recaptcha.rb
```diff
@@ -39,6 +40,10 @@ def self.configure
     yield(config)
   end
 
+  def self.secure_token
```
grosser: no need for a public api I think ... just call the constant directly

grosser:
please rebase and add a tests

astuyve:
@grosser I might be missing something, but it looks like master is failing (and this branch) due to some minitest issues: https://travis-ci.org/ambethia/recaptcha/builds/91877333

grosser:
yeah ... that's me messing around :D 
all green now!

astuyve:
Okay the tests look good. I'm to rebase. FYI @grosser I didn't write a test to make sure it generates a token because you already did here: https://github.com/ambethia/recaptcha/commit/1a151e871df7f3a6c96994549ae07f63c52a3735#diff-35bfa94484a999878842a1178b1216b3R70

lib/recaptcha/token.rb
```diff
@@ -0,0 +1,19 @@
+module Recaptcha
+  module Token
+    require 'json'
```
grosser: I like to keep requires at the top of the file

lib/recaptcha/token.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def self.secure_token
+      private_key  =  Recaptcha.configuration.private_key
+      raise RecaptchaError, "No private key specified." unless private_key
```
grosser: need to require recaptcha to get the error ?

lib/recaptcha/token.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      stoken_json = {'session_id' => SecureRandom.uuid, 'ts_ms' => (Time.now.to_f * 1000).to_i}.to_json
```
grosser: require securerandom

lib/recaptcha/token.rb
```diff
@@ -0,0 +1,19 @@
...
+      cipher = OpenSSL::Cipher::AES128.new(:ECB)
```
grosser: require openssl

lib/recaptcha/token.rb
```diff
@@ -0,0 +1,19 @@
...
+      private_key_digest = Digest::SHA1.digest(private_key)[0...16]
+
+      cipher.encrypt
+      cipher.key = private_key_digest
+      encrypted_stoken = cipher.update(stoken_json) << cipher.final
+      Base64.urlsafe_encode64(encrypted_stoken).gsub(/\=+\Z/, '')
```
grosser: require base64

test/client_helper_test.rb
```diff
@@ -77,6 +77,13 @@
       html = recaptcha_tags(stoken: false)
       html.must_include "<div class=\"g-recaptcha\" data-sitekey=\"0000000000000000000000000000000000000000\"></div>"
     end
+
+    it "should raise an error if secure_token is called without a private_key" do
```
grosser: how about the simpler:
`it "raises if secure_token is called without a private_key" do`

lib/recaptcha/token.rb
```diff
@@ -0,0 +1,19 @@
+module Recaptcha
+  module Token
+    require 'json'
+
+    def self.secure_token
```
grosser: since it's a public api now, a basic test like `it 'generates a token'` would be nice

astuyve: in addition to https://github.com/ambethia/recaptcha/commit/1a151e871df7f3a6c96994549ae07f63c52a3735#diff-35bfa94484a999878842a1178b1216b3R70 ?

grosser: yes, since it is it's own public api ... something simple is good enough ... just want to have a unit test so people know where to add if they change/fix it

lib/recaptcha/token.rb
astuyve: Is there a reason where none of this is required in the existing implementation, but it's required now?

lib/recaptcha/token.rb
astuyve: Gotchya.

lib/recaptcha/token.rb
grosser: used to be 1 big ball and that ball was tested, now it's a new public api that people will use independently, so I'd prefer to have at least a basic smoke test for it ... not really necessary, just out of habit / best practice

astuyve: Okay. Hence the `require recaptcha` comment above

grosser: exactly ... I think that's why you opened this PR, to use it standalone, right ?

astuyve: We're using it in tandem, which in hindsight explains why I blatantly overlooked this. We have a SSO domain which uses `<%= recaptcha_tags %>` (hence the gem), and then a ton of other domains which use just the token.

grosser:
0.6.0!

astuyve:
Thanks!

PR #156, opened 2016-01-27
tests for environment variable set in rack-based apps

fivepapertigers:


grosser:
I'd say put it before Rails detection ... also not sure why but all tests break ... I think travis might have RACK_ENV set ... so unset it in test_helper I guess

fivepapertigers:
Ok, should be good now. Got the tests to pass, looks like travis does set RACK_ENV.

grosser:
thx, 1.1.0 is out!

fivepapertigers:
Thanks for taking care of it quickly! Much appreciated.

PR #166, opened 2016-03-22
Update error messages for reCAPTCHA v2.0

kplattret:
Error messages mentioning "Word verification" are no longer relevant with v2.0 so I suggest to replace them with more generic ones.


grosser:
thx!
1.1.1 is out!

Issue #171, opened 2016-04-04
Make Rails Flash Optional


tomprats:
First I'd like to say I'm really grateful for this gem!

But one of the unexpected things when I was using it is that it seems to add the recaptcha message to the flash if it fails. Since I'm using the model implementation, it doesn't make a lot of sense for the error to show up on the user's page twice.

I was wondering if we could add an optional argument as to whether or not it would add anything to the flash object. I'd definitely be willing to submit a pull request if this is something you're interested in.

Thanks!

grosser:
Ideally it should be smart enough to detect wether the model implementation
is used and then not add the duplicate flash ... but option is fine too if
that is tricky ...

tomprats:
That's what I expected it to be like but that might be a breaking change. I could probably go either way though

grosser:
I'd prefer if we try the 'correct' change first :)
If that is somehow a breaking change we can add a changelog entry or
support weird legacy usecases with an option.

grosser:
1.3.0

PR #172, opened 2016-04-07
Skip flash error when model is present

tomprats:
In reference to https://github.com/ambethia/recaptcha/issues/171. This is all that would be necessary without adding an option. Let me know if you want me to try it with some sort of config option, to support previous behavior

### DIFF
--- a/lib/recaptcha/verify.rb
+++ b/lib/recaptcha/verify.rb
@@ -28,7 +28,7 @@ module Recaptcha
         answer = reply['success']
 
         if hostname_valid?(reply['hostname'], options[:hostname]) && answer.to_s == 'true'
-          flash.delete(:recaptcha_error) if recaptcha_flash_supported?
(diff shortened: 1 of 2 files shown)


lib/recaptcha/verify.rb
```diff
@@ -74,7 +74,7 @@ def hostname_valid?(hostname, validation)
 
     def recaptcha_error(model, attribute, message, key, default)
       message = message || Recaptcha.i18n(key, default)
-      flash[:recaptcha_error] = message if recaptcha_flash_supported?
+      flash[:recaptcha_error] = message if recaptcha_flash_supported? && model.nil?
```
grosser: that would make more sense as an if/else

grosser:
1.3.0 is out ... let's see if anyone complains ... does not make any sense to want both afaik ...

PR #178, opened 2016-05-03
Document the "hl" option for recaptcha_tags

mcls:
More info:
- https://developers.google.com/recaptcha/docs/display#config
- https://developers.google.com/recaptcha/docs/language


grosser:
❤️

PR #182, opened 2016-05-19
Feature/recaptcha check optimization

tuned-up:
Optimization: do not query google if reCaptcha was not submitted

### DIFF
--- a/lib/recaptcha/version.rb
+++ b/lib/recaptcha/version.rb
@@ -1,3 +1,3 @@
 module Recaptcha
-  VERSION = "2.1.0"
+  VERSION = "2.0.0"
 end
--- a/lib/recaptcha/version.rb
+++ b/lib/recaptcha/version.rb
@@ -1,3 +1,3 @@
 module Recaptcha
-  VERSION = "2.0.0"
+  VERSION = "2.1.0"
 end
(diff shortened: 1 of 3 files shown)


lib/recaptcha/version.rb
```diff
@@ -1,3 +1,3 @@
 module Recaptcha
-  VERSION = "2.0.0"
+  VERSION = "2.1.0"
```
grosser: never bump in prs

test/verify_test.rb
```diff
@@ -145,6 +145,25 @@
       assert_nil @controller.flash[:recaptcha_error]
     end
 
+    describe 'when recaptcha response is empty' do
+      before do
+        @controller.params = { 'g-recaptcha-response' => ""}
```
grosser: how does this happen ? `when user did not check the checkbox` ?

lib/recaptcha/version.rb
tuned-up: OK - fixed

test/verify_test.rb
tuned-up: Yes, exactly.

grosser:
FYI refactor https://github.com/ambethia/recaptcha/commit/3f53c5c613470548ba318a0b2fb40a24d6848d17

grosser:
thx again, v2.1.0 is out!

PR #185, opened 2016-05-24
Clean up after with_configuration exception

nate00:
If an exception is raised inside a `with_configuration` block, then the inner configuration "leaks" outside the block. This PR ensures that the cleanup always takes place.


grosser:
❤️  good catch!

grosser:
2.2.0

PR #187, opened 2016-05-26
get rid of all unsecure things

grosser:
@ambethia @nate00  @astuyve @mcls  @JustMikey  @davingee @tomprats  thoughts ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,6 @@ Some of the options available:
 
 | Option      | Description |
 |-------------|-------------|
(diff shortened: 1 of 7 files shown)


astuyve:
Looks like Google has removed all language about choosing between SSL and non-SSL, including references to the option itself. Even the code snippets are `https`:
image

I say 👍

grosser:
need to double check we use async too ...

ambethia:
Full HTTPS sounds great to me. I can't think of a reason in 2016 to not be using HTTPS exclusively. For edge case users, reimplementing the functionality of (or forking) this gem is not a difficult task.

PR #188, opened 2016-06-10
Provides clearer message with action for recaptcha_tags

spilth:
After a recent upgrade I was getting `SSL is always true` all over my Rails 5 feature specs but had no idea what was causing it and how to fix it. I think this clearer error message gives users an action to take to fix it.


ErebusBat:
👍

reneklacan:
👍

rakvat:
:+1:

PR #189, opened 2016-06-12
Unfreeze SKIP_VERIFY_ENV array

gkopylov:
Currently there is an error ```can't modify frozen Array``` when you try to delete ```test``` from ```SKIP_VERIFY_ENV``` array( ```Recaptcha.configuration.skip_verify_env.delete('test')``` ) while testing.

### DIFF
--- a/lib/recaptcha.rb
+++ b/lib/recaptcha.rb
@@ -12,7 +12,6 @@ module Recaptcha
 
   USE_SSL_BY_DEFAULT              = false
   HANDLE_TIMEOUTS_GRACEFULLY      = true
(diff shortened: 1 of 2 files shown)


grosser:
let's make it a class variable then, constants should be erm constant ;)

gkopylov:
@grosser may be it would be better to just delete this constant and set this array explicitly here https://github.com/ambethia/recaptcha/blob/master/lib/recaptcha/configuration.rb#L34 ?

grosser:
that works too!

gkopylov:
@grosser can you merge it?

grosser:
thx, released as 3.2.0

PR #192, opened 2016-07-03
Add id tag attribute and script true/false options

sunchess:
some times you need this two options

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -26,11 +26,12 @@ module Recaptcha
       data_attributes[:sitekey] = public_key
       data_attributes = data_attributes.map { |k, v| %(data-#{k.to_s.tr('_', '-')}="#{v}") }.join(" ")
 
-      data_attributes << %( id="#{options[:id]}") if options[:id]
+      tag_attributes = ""
+      tag_attributes << %(id="#{options[:id]}") if options[:id]
 
       html = ""
-      html << %(<script src="#{script_url}" async defer></script>\n) if options[:script]
-      html << %(<div class="g-recaptcha" #{data_attributes}></div>\n)
+      html << %(<script src="#{script_url}" async defer></script>\n) if options.fetch(:script, true)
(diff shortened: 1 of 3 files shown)


lib/recaptcha/client_helper.rb
```diff
@@ -23,7 +26,10 @@ def recaptcha_tags(options = {})
       data_attributes[:sitekey] = public_key
       data_attributes = data_attributes.map { |k, v| %(data-#{k.to_s.tr('_', '-')}="#{v}") }.join(" ")
 
-      html = %(<script src="#{script_url}" async defer></script>\n)
+      data_attributes << %( id="#{options[:id]}") if options[:id]
+
+      html = ""
+      html << %(<script src="#{script_url}" async defer></script>\n) if options[:script]
```
grosser: `if options.fetch(:script, true)`

test/client_helper_test.rb
```diff
@@ -32,4 +32,18 @@
       recaptcha_tags
     end
   end
+
+  it "including id in tag" do
```
grosser: `it including id in tag` <- what kind of english is that :)

lib/recaptcha/client_helper.rb
```diff
@@ -23,7 +26,10 @@ def recaptcha_tags(options = {})
       data_attributes[:sitekey] = public_key
       data_attributes = data_attributes.map { |k, v| %(data-#{k.to_s.tr('_', '-')}="#{v}") }.join(" ")
 
-      html = %(<script src="#{script_url}" async defer></script>\n)
+      data_attributes << %( id="#{options[:id]}") if options[:id]
```
grosser: it's not a data_attribute, so this looks a bit strange ... add it straight into the html below ?

sunchess:
ok, fixed according notes :)

grosser:
3.3.0 FYI cleanup https://github.com/ambethia/recaptcha/commit/5f01ef4ee695b7faca239863e2235b6ee4e786c6

PR #202, opened 2016-10-26
Update fallback HTML

guilffer:
reCaptcha [documentation](https://developers.google.com/recaptcha/docs/faq#does-recaptcha-support-users-that-dont-have-javascript-enabled) provides a different fallback HTML version for the `<noscript>` tag, it was probably updated recently and it is causing the fallback iframe to not render properly.

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -34,23 +34,24 @@ module Recaptcha
       if options[:noscript] != false
         html << <<-HTML
           <noscript>
-            <div style="width: 302px; height: 352px;">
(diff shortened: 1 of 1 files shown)


grosser:
could you add a before/after screenshot for example from the demo app ?

guilffer:
@grosser sure! 🙂

#### Before
screen shot 2016-10-26 at 15 41 12

#### After
screen shot 2016-10-26 at 15 40 45

grosser:
oh boy ... that looks awful :D

grosser:
0.3.4

PR #204, opened 2016-11-14
Rename api keys to v2 scheme

jacobat:
The Recaptcha v2 API uses the names site_key / secret_key instead of public_key
/ private_key. Make the code reflect this change.


grosser:
nice!

grosser:
4.0.0

PR #207, opened 2016-12-21
Support invisible reCAPTCHA

ryush00:
Recently, they released new reCAPTCHA. (currently beta)

https://developers.google.com/recaptcha/docs/invisible


grosser:
... was expecting some funky PR ... this is simple :D

juampi:
I don't see how this change implements support for invisible reCAPTCHA. It looks like that data-badge attribute is entirely optional. Did the beta API change? Or am I missing something?

PR #211, opened 2017-02-03
Add invisible reCAPTCHA support

juampi:
This pull request adds preliminary support for the [invisible reCAPTCHA](https://developers.google.com/recaptcha/docs/invisible).

It introduces the `invisible_recaptcha_tags` helper, which adds reCAPTCHA's JavaScript resource as well as a submit button. The user also needs to set a callback to handle competition of the captcha using the `callback` option.

More info: https://developers.google.com/recaptcha/docs/invisible#auto_render

### DIFF
--- a/demo/rails/.env
+++ b/demo/rails/.env
@@ -1,3 +1,4 @@
-# reCAPTCHA keys for testing purposes (https://developers.google.com/recaptcha/docs/faq)
-RECAPTCHA_SITE_KEY = 6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
-RECAPTCHA_SECRET_KEY = 6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
+# reCAPTCHA keys for testing purposes
+# Make your own at https://www.google.com/recaptcha
+RECAPTCHA_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
+RECAPTCHA_SECRET_KEY=6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
--- a/demo/rails/.env
+++ b/demo/rails/.env
@@ -1,3 +1,3 @@
-# these will only work on localhost (registered on mg@z) ... make your own at https://www.google.com/recaptcha
-RECAPTCHA_PUBLIC_KEY=6Le7oRETAAAAAETt105rjswZ15EuVJiF7BxPROkY
-RECAPTCHA_PRIVATE_KEY=6Le7oRETAAAAAL5a8yOmEdmDi3b2pH7mq5iH1bYK
+# reCAPTCHA keys for testing purposes (https://developers.google.com/recaptcha/docs/faq)
+RECAPTCHA_SITE_KEY = 6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
+RECAPTCHA_SECRET_KEY = 6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
--- a/demo/rails/app/views/captcha/index.html.erb
+++ b/demo/rails/app/views/captcha/index.html.erb
@@ -38,7 +38,7 @@
   <script src="https://www.google.com/recaptcha/api.js?onload=onloadCallback&render=explicit" async defer></script>
 <% elsif params[:invisible] %>
   <script>
-    function onSubmitCallback() {
+    var onSubmitCallback = function() {
       document.getElementById("invisible-recaptcha-form").submit();
     }
   </script>
--- a/demo/rails/app/views/captcha/index.html.erb
+++ b/demo/rails/app/views/captcha/index.html.erb
@@ -43,7 +43,7 @@
     }
   </script>
   <%= form_tag "/captchas", id: "invisible-recaptcha-form" do %>
-    <%= invisible_recaptcha_tags callback: 'onSubmitCallback' %>
+    <%= invisible_recaptcha_tags callback: 'onSubmitCallback', text: 'Save changes' %>
   <% end %>
 <% else %>
   <%= form_tag "/captchas" do %>
--- a/demo/rails/app/views/captcha/index.html.erb
+++ b/demo/rails/app/views/captcha/index.html.erb
@@ -36,11 +36,21 @@
     <%= submit_tag %>
   <% end %>
   <script src="https://www.google.com/recaptcha/api.js?onload=onloadCallback&render=explicit" async defer></script>
-  <%= link_to 'Single ?', '?' %>
+<% elsif params[:invisible] %>
+  <script>
+    function onSubmitCallback() {
+      document.getElementById("invisible-recaptcha-form").submit();
+    }
+  </script>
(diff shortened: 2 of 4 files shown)


demo/rails/.env
```diff
@@ -1,3 +1,3 @@
-# these will only work on localhost (registered on mg@z) ... make your own at https://www.google.com/recaptcha
-RECAPTCHA_PUBLIC_KEY=6Le7oRETAAAAAETt105rjswZ15EuVJiF7BxPROkY
-RECAPTCHA_PRIVATE_KEY=6Le7oRETAAAAAL5a8yOmEdmDi3b2pH7mq5iH1bYK
+# reCAPTCHA keys for testing purposes (https://developers.google.com/recaptcha/docs/faq)
+RECAPTCHA_SITE_KEY = 6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
```
grosser: I'd rather not mess with the spacing here ...

demo/rails/.env
```diff
@@ -1,3 +1,3 @@
-# these will only work on localhost (registered on mg@z) ... make your own at https://www.google.com/recaptcha
-RECAPTCHA_PUBLIC_KEY=6Le7oRETAAAAAETt105rjswZ15EuVJiF7BxPROkY
-RECAPTCHA_PRIVATE_KEY=6Le7oRETAAAAAL5a8yOmEdmDi3b2pH7mq5iH1bYK
+# reCAPTCHA keys for testing purposes (https://developers.google.com/recaptcha/docs/faq)
```
grosser: nobody reads faqs ... the signup site was nicer imo

demo/rails/app/views/captcha/index.html.erb
```diff
@@ -36,11 +36,21 @@
     <%= submit_tag %>
   <% end %>
   <script src="https://www.google.com/recaptcha/api.js?onload=onloadCallback&render=explicit" async defer></script>
-  <%= link_to 'Single ?', '?' %>
+<% elsif params[:invisible] %>
+  <script>
+    function onSubmitCallback() {
```
grosser: can't have a global method with such a generic name ... also should not use inline js to be CSP compliant

test/client_helper_test.rb
```diff
@@ -46,4 +46,31 @@
...
+    it "raises without site key" do
+      Recaptcha.configuration.site_key = nil
+      assert_raises Recaptcha::RecaptchaError do
+        invisible_recaptcha_tags
+      end
+    end
+
+    it "includes id as button attribute" do
+      html = invisible_recaptcha_tags(id: 'my_id')
+      html.must_include(
```
grosser: don't need multiline here

lib/recaptcha/client_helper.rb
```diff
@@ -60,5 +60,25 @@ def recaptcha_tags(options = {})
 
       html.respond_to?(:html_safe) ? html.html_safe : html
     end
+
+    # Invisible reCAPTCHA implementation
+    def invisible_recaptcha_tags(options = {})
+      site_key = options[:site_key] || Recaptcha.configuration.site_key!
+      script_url = Recaptcha.configuration.api_server_url
+      data_attributes = [:badge, :theme, :type, :callback, :expired_callback, :size, :tabindex]
```
grosser: don't repeat this .. use a constant

lib/recaptcha/client_helper.rb
```diff
@@ -60,5 +60,25 @@ def recaptcha_tags(options = {})
...
+      data_attributes = options.each_with_object({}) do |(k, v), a|
+        a[k] = v if data_attributes.include?(k)
+      end
+      data_attributes[:sitekey] = site_key
+      tag_attributes = data_attributes.map { |k, v| %(data-#{k.to_s.tr('_', '-')}="#{v}") }.join(" ")
```
grosser: same here, more copy-paste

demo/rails/.env
juampi: Alright. The reason I changed the keys was because the old ones were not working at all. And those are "official" ones meant for testing purposes.

demo/rails/.env
juampi: Sure. I put the link to show where I got the keys from.

grosser:
ah, then write test credentials from xxx :D
... nobody needs to go there and double-check

On Thu, Feb 2, 2017 at 9:42 PM, Juan Pablo Rinaldi <notifications@github.com

demo/rails/app/views/captcha/index.html.erb
juampi: Not sure what you mean with being CSP compliant. Isn't there more inline JavaScript on this file already? And would making an anonymous function and storing it in a `var` be enough to deal with the generic name issue?

grosser: nvm ... I though this was part of the generated code ... fine as it is :D

lib/recaptcha/client_helper.rb
```diff
@@ -1,35 +1,43 @@
...
-    # using the Configuration.
-    def recaptcha_tags(options = {})
-      if options.key?(:stoken)
-        raise(RecaptchaError, "Secure Token is deprecated. Please remove 'stoken' from your calls to recaptcha_tags.")
-      end
-      if options.key?(:ssl)
-        raise(RecaptchaError, "SSL is now always true. Please remove 'ssl' from your calls to recaptcha_tags.")
-      end
-
+    def build_html(options = {})
```
grosser: this is included everywhere, so needs a less generic name ... preferably starting with recaptcha_

lib/recaptcha/client_helper.rb
```diff
@@ -1,35 +1,43 @@
...
+      tag_attributes << %( class="g-recaptcha #{options[:class]}")
 
       html = ""
       html << %(<script src="#{script_url}" async defer></script>\n) if options.fetch(:script, true)
-      html << %(<div class="g-recaptcha" #{tag_attributes}></div>\n)
+      html
+
+      fallback_uri = "#{script_url.chomp('.js')}/fallback?k=#{site_key}"
+
+      return html, tag_attributes, fallback_uri
```
grosser: that return value does not make much sense ... how about a hash ?

lib/recaptcha/client_helper.rb
```diff
@@ -60,5 +68,12 @@ def recaptcha_tags(options = {})
 
       html.respond_to?(:html_safe) ? html.html_safe : html
     end
+
+    # Invisible reCAPTCHA implementation
+    def invisible_recaptcha_tags(options = {})
+      html, tag_attributes = build_html(options)
+      html << %(<button type="submit" #{tag_attributes}>#{options[:text]}</button>\n)
+      html.respond_to?(:html_safe) ? html.html_safe : html
+    end
```
grosser: much better :D

lib/recaptcha/client_helper.rb
juampi: I see. I'm no Ruby expert, but would making it private help?

grosser:
sadly no, since it is included into global helpers ...
can either namespace or make it a static method (def self.) and then call
it like Recatcha::ClientHelper.build_html

grosser:
4.1.0

grosser:
... arg some docs maybe ? :D

Issue #214, opened 2017-03-09
Client connecting through ruby application to recaptcha services


wyhasany:
My web application uses your gem. It is located inside local network and only this host in LAN has access to the Internet. I would like to have a configuration which rewrite scripts to use reverse proxy instead of google web pages. Is it hard to implement that e of functionality?

grosser:
Try proxy https://github.com/ambethia/recaptcha#recaptchaconfigure

... alternatively changing verify_url / api_server_url might work

wyhasany:
This property works only on the server side not on the client.

wyhasany:
I need to edit: 'server_url' => 'https://www.google.com/recaptcha/api.js', from recaptcha.rb

grosser:
make a PR to change it to be configureable (for example like `skip_verify_env`) ... we don't need the CONFIG object anymore

sushengloong:
I created PR #221 to address this issue. For your consideration :smile:

grosser:
PR looks good, should be able to merge that shortly ... closing this

PR #221, opened 2017-04-26
Make api_server_url and verify_url configurable

sushengloong:
This is so that the URL can be set to some proxies instead of the
hardcoded Google links.
This commit is inspired by https://github.com/ambethia/recaptcha/issues/214\#issuecomment-285257412

### DIFF
--- a/lib/recaptcha/configuration.rb
+++ b/lib/recaptcha/configuration.rb
@@ -37,9 +37,6 @@ module Recaptcha
 
       @secret_key           = ENV['RECAPTCHA_SECRET_KEY']
       @site_key             = ENV['RECAPTCHA_SITE_KEY']
-
-      @api_server_url       = ENV['RECAPTCHA_API_SERVER_URL']
-      @verify_url           = ENV['RECAPTCHA_VERIFY_URL']
     end
 
     def secret_key!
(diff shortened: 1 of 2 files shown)


lib/recaptcha/configuration.rb
```diff
@@ -28,14 +28,18 @@ module Recaptcha
...
+      :handle_timeouts_gracefully, :hostname
 
     def initialize #:nodoc:
       @skip_verify_env            = %w[test cucumber]
       @handle_timeouts_gracefully = HANDLE_TIMEOUTS_GRACEFULLY
 
       @secret_key           = ENV['RECAPTCHA_SECRET_KEY']
       @site_key             = ENV['RECAPTCHA_SITE_KEY']
+
+      @api_server_url       = ENV['RECAPTCHA_API_SERVER_URL']
```
grosser: set it in CONFIG, since this does not need to be changed at runtime

grosser: ... or maybe leave it to the client to configure that ?

sushengloong: Sorry, I don't quite understand. Actually I tried to follow how `@secret_key` and `@site_key` are set, except that if client never configures the url (or set via `ENV`), it will fallback to the CONFIG. Can you please explain further on how I can improve this?

grosser: I was thinking about just not setting them here since very few users will need it ... and those that do can configure the client manually

PR #223, opened 2017-04-28
Allow random attributes to be rendered for invisible captcha button

### LINKED ISSUE #222: Use Rails button_tag helper for invisible captcha
Just wondering if there's a reason to not use button_tag helper for invisible recaptcha? I'm asking because I need to set more attributes on the generated button tag (e.g. disabled and other data-X attributes). With the way it's setup now, you can only specify class it seems.

ngan:
Fixes #222

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -10,10 +10,12 @@ module Recaptcha
         raise(RecaptchaError, "SSL is now always true. Please remove 'ssl' from your calls to recaptcha_tags.")
       end
 
+      noscript = options.delete(:noscript)
+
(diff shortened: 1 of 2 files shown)


grosser:
hmm this escalated :D 
how about `html: {foo: bar}` so we don't have to do delete ?

ngan:
Won't that change method signatures, resulting in breaking changes?

ngan:
If that's ok, then I have even more improvements to make.

grosser:
hmmm ... this is all mixed up atm :/

grosser:
yeah ... I think what you did here is pretty good ... users will see that an option was ignored when it ends up in the tag ... 
we might have to later change this if we want an option that is also a html attribute ... but at least we don't need to break anyones code just yet ...

grosser:
4.3.0

PR #227, opened 2017-05-16
correctly generate data-expired-callback attribute name

datanoise:
Specifying the `:expired_callback` attribute to `recaptcha_tags` incorrectly generates `data-expired_callback` name instead of `data-expired-callback`.


grosser:
looks good, but needs a test or I will break it in the next release :trollface:

PR #231, opened 2017-07-11
Update README.md to have alternative config use correct names?

tansaku:
I was just following the instructions and got this error:

```
config/initializers/recaptcha.rb:2:in `block in <top (required)>': undefined method `site_key=' for #<Recaptcha::Configuration:0x007f9606ad9a78> (NoMethodError)
Did you mean?  private_key=
```
Seems like the config names have changed - I've updated them to ones that work in this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,8 +155,8 @@ Recaptcha.configuration.skip_verify_env.delete("test")
 ```Ruby
 # config/initializers/recaptcha.rb
 Recaptcha.configure do |config|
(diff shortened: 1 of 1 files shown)


grosser:
hehe ... no that's wrong ... you got an old version of recaptcha installed so the docs don't work ...

tansaku:
@grosser the version that installed for us was    recaptcha (0.4.0)

tansaku:
gosh that really is very old - wonder why we got that one ... great that it actually works - will have to upgrade I guess ...

PR #234, opened 2017-09-01
Fix invisible_recaptcha_tags to work in testing modes

rickychilcott:
Add relevant javascript and markup options to invisible_recaptcha_tags ONLY we are not in a testing mode

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -60,7 +60,7 @@ module Recaptcha
       attributes = {}
       fallback_uri = ""
 
-      attributes["class"] = %(g-recaptcha #{options.delete(:class)})
+      attributes["class"] = "g-recaptcha #{options.delete(:class)}"
 
       unless Recaptcha::Verify.skip?(options[:env])
         hl = options.delete(:hl).to_s
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -61,20 +61,21 @@ module Recaptcha
         class: ["g-recaptcha", options.delete(:class)].join(" ")
       }
 
-      hl = options.delete(:hl)
-      script_url = Recaptcha.configuration.api_server_url
-      script_url += "?hl=#{hl}" unless hl.to_s == ""
-      html << %(<script src="#{script_url}" async defer></script>\n) unless options.delete(:script) == false
-      fallback_uri = "#{script_url.chomp('.js')}/fallback?k=#{site_key}"
+      unless Recaptcha::Verify.skip?(options[:env])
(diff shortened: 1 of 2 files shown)


rickychilcott:
I ran into a bug when I tried to run (previously passing) tests after adding the `recaptcha` gem and specifically using the `invisible_recaptcha_tags` method. This PR cleans up the markup so that it only passes through the needed details to get the UI to look the same, but not load the non-test related javascript.

Thanks for your work on this gem @ambethia!

rickychilcott:
Also, let me know if you want a test for this. I just wrap the one section with an if clause, but I'm willing to prove it works.

grosser:
looks reasonable to me ... a test would be nice though ...

lib/recaptcha/client_helper.rb
```diff
@@ -57,24 +57,27 @@ def invisible_recaptcha_tags(options = {})
...
-      }
+      attributes = {}
+      fallback_uri = ""
 
-      hl = options.delete(:hl)
-      script_url = Recaptcha.configuration.api_server_url
-      script_url += "?hl=#{hl}" unless hl.to_s == ""
-      html << %(<script src="#{script_url}" async defer></script>\n) unless options.delete(:script) == false
-      fallback_uri = "#{script_url.chomp('.js')}/fallback?k=#{site_key}"
+      attributes["class"] = %(g-recaptcha #{options.delete(:class)})
```
grosser: can just use quotes instead of `%(`

test/client_helper_test.rb
```diff
@@ -77,5 +77,11 @@
       html = invisible_recaptcha_tags
       html.must_include(" data-sitekey=\"#{Recaptcha.configuration.site_key}\"")
     end
+
+    it "doesn't render script tag when in test mode" do
+      html = invisible_recaptcha_tags(env: "test")
+      html.wont_include("<script")
+      html.wont_include(" data-sitekey=\"#{Recaptcha.configuration.site_key}\"")
```
grosser: can make this more generic `"data-sitekey"`

test/client_helper_test.rb
```diff
@@ -77,5 +77,11 @@
       html = invisible_recaptcha_tags
       html.must_include(" data-sitekey=\"#{Recaptcha.configuration.site_key}\"")
     end
+
+    it "doesn't render script tag when in test mode" do
```
grosser: more like "when verification is disabled" ?

grosser:
sweet!

rickychilcott:
Thanks for the review @grosser!

grosser:
4.4.0 ... hope nobody comes and complains about their "usecase" :D

PR #235, opened 2017-09-08
Bug fix: :hl option should not alter api_server_url value

konstantin-dzreev:
otherwise the url becomes insane with many '?hl=<locale>' params


grosser:
nice, even tests :D

konstantin-dzreev:
hehe... Travis did not think so on the first build ;)

grosser:
here comes 4.4.1  :)

PR #237, opened 2017-09-15
Update json in Gemfile.lock for ruby 2.4+

philipqnguyen:
1.8.3 does not work with ruby 2.4+


philipqnguyen:
see here: https://github.com/flori/json/issues/303

Issue #220, opened 2017-04-20
improvements for `invisible_recaptcha_tags`?


philipqnguyen:
Thanks @juampi for originally implementing `invisible_recaptcha_tags`! I was able to successfully get it to work on my app.

Now, I'm wondering if `invisible_recaptcha_tags` public api has diverged too much from the original intention?

The two things you must do differently from `recaptcha_tags` are:

1. You must specify a callback function, which would be called after verification with Google's reCAPTCHA service. Ideally, this callback function would submit the actual form.

2. The invisible_recaptcha_tags will generate a submit button for you. You should not need to add a f.submit or submit_tag to your form.

Do you think we should automate the callback of the form submission and to remove the generation of the submit button?

grosser:
I think it's easy enough ... automating JS is rarely fun ... 
We could add a `submit: '#foobar'` option to it that then generates the JS inline, but it shouldalways be an option for cases where js injection is disabled.

philipqnguyen:
What do you think about adding a default for when there is no option passed?

Here is something that has come to my mind:
```javascript
// app/assets/javascripts/application.js
//= require recaptcha
```

```erb
<%= form_for @foo do |f| %>
  <!-- other form fields -->
  <%= invisible_recaptcha_tags %> <!-- no options passed, no button generation. -->
  <%= f.submit 'Save' %>
<% end %>
```
The recaptcha.js would be a new file included in the gem. It would contain some js functions to stop the form on submit, wait for Google's reCAPTCHA verification response, and then it resumes the submission.

Just an idea.

grosser:
sounds good ... maybe make it `require recaptcha/invisible` so it's clear what it is for

philipqnguyen:
I can try to work on this on my free time. Though if someone else wants to take a stab, be my guest.

chrismd3:
I'd really like to see this. I was looking for a way to do something like this with the existing tag already.

Superpencil:
This helped me a lot as a first time user, seeing that empty thing there, thanks!

philipqnguyen:
I'll close this issue now since #238 pretty much solved the pain that this issue was surfacing.

PR #238, opened 2017-09-16
Add default callback for invisible recaptcha #220

philipqnguyen:
~This is a work in progress commit and looking for some ideas.
It is working towards the first part of #220.~

~This adds a default callback that calls the function
`invisibleRecaptchaSubmit` if no callback was passed to
`invisible_recaptcha_tags`~

~The drawback to this approach is that it won't work if there are
multiple recaptcha forms.~

Add default callback for invisible recaptcha #220
    
This adds a default callback that calls the function
`invisibleRecaptchaSubmit` if no callback was passed to
`invisible_recaptcha_tags`. The function, `invisibleRecaptchaSubmit`,
uses the `closest()` method to find the closest form to the
`.g-recaptcha` element. However, not all browsers support `closest()`,
so in that case, a polyfill of `closest` is executed. The polyfill is
based on MDN's polyfill suggestion which should support IE9+
https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill
    
The drawback to this approach is that it won't work if there are
multiple recaptcha forms. If you want to use invisible_recaptcha_tags
with multiple forms, you would still need to use custom callbacks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,25 @@ Some of the options available:
 
 ## invisible_recaptcha_tags
 
-[Invisible reCAPTCHA](https://developers.google.com/recaptcha/docs/invisible) has two differences:
+Make sure to read [Invisible reCAPTCHA](https://developers.google.com/recaptcha/docs/invisible).
 
-1. Needs a callback function, which is called after verification with Google's reCAPTCHA service. This callback function must submit the form.
-2. The `invisible_recaptcha_tags` generates a submit button.
+### With a single form on a page
+
+1. The `invisible_recaptcha_tags` generates a submit button for you.
+
+```Erb
+<%= form_for @foo do |f| %>
+  # ... other tags
+  <%= invisible_recaptcha_tags text: 'Submit form' %>
+<% end %>
+```
+
+Then, add `verify_recaptcha` to your controller as seen [above](#rails-installation).
+
+### With multiple forms on a page
+
+1. You will need a custom callback function, which is called after verification with Google's reCAPTCHA service. This callback function must submit the form. Optionally, `invisible_recaptcha_tags` currently implements a JS function called `invisibleRecaptchaSubmit` that is called when no `callback` is passed. Should you wish to override `invisibleRecaptchaSubmit`, you will need to use `invisible_recaptcha_tags script: false`, see lib/recaptcha/client_helper.rb for details.
+2. The `invisible_recaptcha_tags` generates a submit button for you.
 
 ```Erb
 <%= form_for @foo, html: {id: 'invisible-recaptcha-form'} do |f| %>
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -47,9 +47,10 @@ module Recaptcha
 
     # Invisible reCAPTCHA implementation
     def invisible_recaptcha_tags(options = {})
+      options = {callback: 'invisibleRecaptchaSubmit'}.merge options
       text = options.delete(:text)
-      html, tag_attributes = Recaptcha::ClientHelper.recaptcha_components(options)
-
+      html, tag_attributes = Recaptcha::ClientHelper.recaptcha_components(options.dup)
+      html << recaptcha_default_callback if recaptcha_default_callback_required?(options)
       html << %(<button type="submit" #{tag_attributes}>#{text}</button>\n)
       html.respond_to?(:html_safe) ? html.html_safe : html
     end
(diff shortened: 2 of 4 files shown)


grosser:
looks interesting ... should add a way to opt-out too ?

philipqnguyen:
Good call. Here is what I'm thinking:

If a callback option is passed, we don't append the js script. Make sense?

grosser:
yep

philipqnguyen:
@grosser 

I amended the previous commits and removed WIP. What do you think?

README.md
```diff
@@ -101,10 +101,29 @@ Some of the options available:
 
 ## invisible_recaptcha_tags
 
-[Invisible reCAPTCHA](https://developers.google.com/recaptcha/docs/invisible) has two differences:
+Make sure to read [Invisible reCAPTCHA](https://developers.google.com/recaptcha/docs/invisible).
+
+### With a single form on a page
+
+There is one thing you need to take notice:
```
grosser: redundant line ?

lib/recaptcha/client_helper.rb
```diff
@@ -84,5 +85,52 @@ def self.recaptcha_components(options = {})
...
+    private
+
+    def default_callback_name
+      'invisibleRecaptchaSubmit'
+    end
+
+    def default_callback_script
+      <<-HTML
+        <script>
+          var createClosestPolyfill = function () {
```
grosser: not a fan of adding global methods ...

lib/recaptcha/client_helper.rb
```diff
@@ -84,5 +85,52 @@ def self.recaptcha_components(options = {})
...
+                    if (ancestor.matches(s)) return ancestor;
+                    ancestor = ancestor.parentElement;
+                } while (ancestor !== null);
+                return null;
+            };
+          };
+          var #{default_callback_name} = function () {
+            var ele = document.getElementsByClassName('g-recaptcha');
+            if (ele.length > 0) {
+              if (!Element.prototype.closest) {
```
grosser: do something simpler like iterating parents until one is a form ?

lib/recaptcha/client_helper.rb
```diff
@@ -84,5 +85,52 @@ def self.recaptcha_components(options = {})
...
+              } else {
+                ele[0].closest('form').submit();
+              }
+            }
+          };
+        </script>
+      HTML
+    end
+
+    def default_callback_required?(options)
```
grosser: all methods are in the global context, so they should be as few as possible and namespaces with recaptcha

lib/recaptcha/client_helper.rb
```diff
@@ -84,5 +85,52 @@ def self.recaptcha_components(options = {})
...
+                } while (ancestor !== null);
+                return null;
+            };
+          };
+          var #{default_callback_name} = function () {
+            var ele = document.getElementsByClassName('g-recaptcha');
+            if (ele.length > 0) {
+              if (!Element.prototype.closest) {
+                createClosestPolyfill();
+                ele[0].closest('form').submit();
```
grosser: duplicated

README.md
philipqnguyen: Probably =)

lib/recaptcha/client_helper.rb
philipqnguyen: Noted. I'll look into nesting this under the default callback function.

lib/recaptcha/client_helper.rb
philipqnguyen: 👍

lib/recaptcha/client_helper.rb
philipqnguyen: 👍

lib/recaptcha/client_helper.rb
philipqnguyen: Nice catch

philipqnguyen:
Good feedback. I'll fix it up some more over the next few days.

philipqnguyen:
I amended those commits with your suggestions.

README.md
```diff
@@ -101,10 +101,25 @@ Some of the options available:
...
+  # ... other tags
+  <%= invisible_recaptcha_tags text: 'Submit form' %>
+<% end %>
+```
+
+Then, add `verify_recaptcha` to your controller as seen [above](#rails-installation).
+
+### With multiple forms on a page
+
+1. You will need a callback function, which is called after verification with Google's reCAPTCHA service. This callback function must submit the form.
```
grosser: add something like this ?

build your own `invisibleRecaptchaSubmit` function and use `invisible_recaptcha_tags script: false`, see lib/recaptcha/client_helper.rb for details.

philipqnguyen: I'm not sure why they need `script: false`? Based on this PR's implementation, they just need to pass in the name of their custom callback function and it will override the default callback.

So it would be more like: 
build your own invisibleRecaptchaSubmit function and use `invisible_recaptcha_tags callback: custom_callback` see lib/recaptcha/client_helper.rb for details.

Unless there was another reason for `script: false`?

grosser: if someone implements `invisibleRecaptchaSubmit` and passes `callback: 'invisibleRecaptchaSubmit'` then the default script would be added again ... so I was thinking tell users to submit `script: false`

philipqnguyen: Ah ok. Make sense. =)

philipqnguyen:
I had to amend this commit d9228993c665e9e71433bf67d9f6d125915a1b44, to update the commit message. It had some old commit message that was not reflective of the recent changes during the course of this PR.

I also updated the readme to include overriding the default `invisible_recaptcha_submit` function by specifying `script: false`

grosser:
4.6.0

PR #240, opened 2017-10-06
Skip optional remote ip when posting

chaman:
The `remoteip` to verify captcha is optional as per https://developers.google.com/recaptcha/docs/verify. 

Adding an option to skip the remoteip when necessary

### DIFF
--- a/lib/recaptcha/verify.rb
+++ b/lib/recaptcha/verify.rb
@@ -63,7 +63,6 @@ module Recaptcha
 
     def recaptcha_verify_via_api_call(request, recaptcha_response, options)
       secret_key = options[:secret_key] || Recaptcha.configuration.secret_key!
-      
 
       verify_hash = {
         "secret"    => secret_key,
--- a/lib/recaptcha/verify.rb
+++ b/lib/recaptcha/verify.rb
@@ -63,17 +63,22 @@ module Recaptcha
 
     def recaptcha_verify_via_api_call(request, recaptcha_response, options)
       secret_key = options[:secret_key] || Recaptcha.configuration.secret_key!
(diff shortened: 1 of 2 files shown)


lib/recaptcha/verify.rb
```diff
@@ -63,19 +63,24 @@ def self.skip?(env)
...
-      remote_ip = (request.respond_to?(:remote_ip) && request.remote_ip) || (env && env['REMOTE_ADDR'])
+      
 
       verify_hash = {
         "secret"    => secret_key,
-        "remoteip"  => remote_ip.to_s,
         "response"  => recaptcha_response
       }
 
+      verify_hash.merge!( "remoteip"  => remote_ip.to_s ) unless options[:skip_remote_ip]
```
grosser: `verify_hash["remoteip"] = remote_ip.to_s`

test/verify_test.rb
```diff
@@ -247,4 +255,11 @@ def expect_http_post(secret_key: Recaptcha.configuration.secret_key)
       "https://www.google.com/recaptcha/api/siteverify?remoteip=1.1.1.1&response=string&secret=#{secret_key}"
     )
   end
+
+  def expect_http_post_without_ip(secret_key: Recaptcha.configuration.secret_key)
```
grosser: does not need to be a method when it's only used once

grosser:
thx, looks good!
... any reason you don't want to send the remote_ip ?

chaman:
just for security reason. It would be ok to send `remote_ip` of bots but we don't want to send valid users ip.

chaman:
@grosser  oops was missing the `request` param for remote_ip method.

grosser:
4.5.0 🎉

Issue #241, opened 2017-10-16
localhost:3000 to create recaptcha site key


rahul2240:
While using this gem on localhost:3000, i was unable to use localhost:3000 has a domain in google recaptcha so i think in readme, this should be mention that localhost will work as a domain but not localhost:3000 while trying to create recaptcha site key

grosser:
can you make a PR to add the explanation to the readme ?

rahul2240:
Yes, will do it 😄

PR #242, opened 2017-10-16
Update readme to create site key for localhost:3000

rahul2240:
If user is working on localost:3000 then localhost will work as a valid domain because localhost:3000 as a domain violates the valid scheme
closes https://github.com/ambethia/recaptcha/issues/241


grosser:
thx!

PR #245, opened 2017-10-19
Include `data-tabindex` attribute

jesse:
This PR pulls out the `tabindex` data attribute, as documented here https://developers.google.com/recaptcha/docs/display


grosser:
thx! v4.6.2

PR #250, opened 2017-12-29
Ignore site key for skipped environments

juampi:
Currently, a site key is required even for environments where verification is skipped. It is inconvenient to have to define one when it serves no actual purpose.

This PR removes that requirement.


grosser:
v4.6.3

PR #253, opened 2018-01-31
added title attribute to iframe for screen readers

bluemango:
Screen reader testers log an error on the Recaptcha iframe because it doesn't have a title


grosser:
thx, 4.6.4

PR #256, opened 2018-02-27
Dasherise the `data-expired-callback` attribute name properly.

Skudo:
Calling `recaptcha_tags` (or `invisible_recaptcha_tags`) with the `expired_callback` argument generated a `data-expired_callback` attribute that does not achieve anything. This fixes that behaviour.

Essentially adds tests for #227.


grosser:
it's supposed to work without rails too ... can you use a static method for this ?

grosser:
thx! 4.6.5

PR #258, opened 2018-03-05
Prevent leaking recaptcha client_helper options as HTML attributes

### LINKED ISSUE #255: invisible_recaptcha_tags emits invalid HTML in a skipped environment
In test environment, `invisible_recaptcha_tags` emits this code:

```html
<button type="submit" class="g-recaptcha " callback="invisibleRecaptchaSubmit">
```

But `callback` is not a valid attribute of the `button` element. In production, it correctly emits `data-callback` instead. 

In `client_helper.rb`, you can see that in skipped environments, the incoming recaptcha params are no longer deleted from the options hash: 

```ruby
      # client_helper.rb line 66
      unless Recaptcha::Verify.skip?(options[:env])
        site_key = options.delete(:site_key) || Recaptcha.configuration.site_key!
        hl = options.delete(:hl).to_s
        script_url += "?hl=#{hl}" unless hl == ""
        html << %(<script src="#{script_url}" async defer></script>\n) unless options.delete(:script) == false
        fallback_uri = %(#{script_url.chomp(".js")}/fallback?k=#{site_key})

        # Pull out reCaptcha specific data attributes.
        [:badge, :theme, :type, :callback, :expired_callback, :size, :tabindex].each do |data_attribute|
          value = options.delete(data_attribute)

          attributes["data-#{data_attribute}"] = value if value
        end

        attributes["data-sitekey"] = site_key
      end

      # Append whatever that's left of options to be attributes on the tag.
      tag_attributes = attributes.merge(options).map { |k, v| %(#{k}="#{v}") }.join(" ")
```

But that means that in the last line, any of those options (badge, theme, type, callback, expired_callback, size, tabindex) just get added to `tag_attributes`. 

And `invisible_recaptcha_tags` always includes the `callback` option: 

```ruby
    # client_helper.rb line 50
    def invisible_recaptcha_tags(options = {})
      options = {callback: 'invisibleRecaptchaSubmit'}.merge options
```

rdj:
Fixes #255 

In addition to the specific issue with `callback` called out in #255, all kinds of ruby options could leak out into the HTML. For example, if the `env: 'test'` option was provided, the HTML would include `env="test"`. This commit ensures that `recaptcha_components` always consumes all its options so they won't leak into the HTML. 

It would likely be better design to pass desired HTML attributes in a separate Hash instead of interleaving them with ruby options, but that is beyond the scope of this commit. 

I sought to keep behavior as similar as possible, specifically: 
* `data-*` recaptcha-related attributes are not emitted in skipped environments (before they were emitted incorrectly, without the `data-` prefix)
* `class="g-recaptcha"` is emitted even in skipped environments

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -50,7 +50,7 @@ module Recaptcha
     def invisible_recaptcha_tags(options = {})
       options = {callback: 'invisibleRecaptchaSubmit'}.merge options
       text = options.delete(:text)
-      html, tag_attributes = Recaptcha::ClientHelper.recaptcha_components(options.dup)
+      html, tag_attributes = Recaptcha::ClientHelper.recaptcha_components(options)
       html << recaptcha_default_callback if recaptcha_default_callback_required?(options)
       html << %(<button type="submit" #{tag_attributes}>#{text}</button>\n)
       html.respond_to?(:html_safe) ? html.html_safe : html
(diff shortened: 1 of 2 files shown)


grosser:
thx, nice refactor :)

4.6.6

PR #259, opened 2018-03-07
Add ability to temporarily disable recaptcha on the fly through an environment variable

rbclark:
It is very useful to be able to turn off reCaptcha through just setting an environment variable. In my case this is useful for developing on a corporate network where it is not possible to hit reCaptcha servers, while still allowing other developers on different networks to be able to test with reCaptcha test keys.

### DIFF
--- a/lib/recaptcha/configuration.rb
+++ b/lib/recaptcha/configuration.rb
@@ -28,7 +28,7 @@ module Recaptcha
   #   end
   #
   class Configuration
-    attr_accessor :skip_verify_env, :secret_key, :site_key, :api_server_url, :verify_url, :proxy,
+    attr_accessor :skip_verify_env, :secret_key, :site_key, :disable, :api_server_url, :verify_url, :proxy,
       :handle_timeouts_gracefully, :hostname
 
     def initialize #:nodoc:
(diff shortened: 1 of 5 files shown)


grosser:
- env vars should start with RECAPTCHA_
 - env vars should be documented in the place where the other env vars are documented
 - needs tests
 - how about using `skip_verify_env << Rails.env if ENV["DISABLE_RECAPTCHA"]` instead ?

rbclark:
@grosser I believe I have addressed all of your bullet points in the last PR. The one spot where I ran into a bit of an issue was I did not want to make this fix rails specific which your last bullet point would've done. I hope my workaround for that is acceptable.

lib/recaptcha/configuration.rb
```diff
@@ -37,6 +37,8 @@ def initialize #:nodoc:
 
       @secret_key           = ENV['RECAPTCHA_SECRET_KEY']
       @site_key             = ENV['RECAPTCHA_SITE_KEY']
+      # Calling @disable will not hit the setter below
+      self.disable          = ['1', 'true', 'yes'].include? ENV['RECAPTCHA_DISABLE']
```
grosser: maybe simpler to just set `@disabled` instead of extra logic to set and remove an env ... which is also kinda unexpected

my previous comment was to suggest not doing this PR and conditionally adding skip_verify_env from the host app, not doing it here ... sorry for the confusion

rbclark:
Sorry, I didn't quite understand your original comment I guess. I have gone ahead and just added `skip_verify_env << Rails.env if ENV["DISABLE_RECAPTCHA"]` to my project instead.

PR #262, opened 2018-03-23
Ruby 1.8.7 and Rails 2.3 compatible

### LINKED ISSUE #261: Backport to Ruby 1.8.7, Rails 2.3?
I have an ancient app that used an old version, needed to upgrade it, so I backported this gem to work on Rails 2.3 and Ruby 1.8.7 (enterprise edition)

https://github.com/PRX/recaptcha/tree/backport_ruby187_rails23

I suspect you don't want these changes in master, but thought I would add an issue here for any fellow travelers to find, and offer to provide a PR for this support in a branch if anyone was interested.

kookster:
* Update ruby version and set gems to 1.8.7 compatible versions

* Jettisoned rubocop and bump, remove from Rakefile

* Add back in railtie.rb, use rails.rb for basic loading

* Use `Rack::Utils.build_query` instead of ruby 1.9+ `URI.encode_www_form`

* Fix all the hash rockets

* Remove `require_relative`, update `maxitest` around to use `minitest-hooks`

* Keep the same gem version as master, but with .r187 suffix

fixes #261


lib/recaptcha/version.rb
```diff
@@ -1,3 +1,3 @@
 module Recaptcha
-  VERSION = "4.6.6".freeze
+  VERSION = "4.6.6.r187".freeze
```
grosser: maybe call it ruby187 to make it more obvious

grosser:
looks great!

change the .travis.yml to also run builds for rails23 and kill all the combinations it should not run ... maybe we even get this green :D

kookster:
I specified ree-1.8.7-2012.02 b/c that's what the old app used, but seems to take travis a dog's age to install that...

kookster:
How about that @grosser - got ourselves a green light!

grosser:
great job!

grosser:
released v4.6.6.ruby187

PR #264, opened 2018-03-28
Allow setting nonce attribute for reCAPTCHA API script

NARKOZ:
This PR adds a nonce attribute to script tag, which can be set by passing a `nonce` option.

This is necessary for setting CSP (Content-Security-Policy) header.

See: https://developers.google.com/recaptcha/docs/faq#im-using-content-security-policy-csp-on-my-website-how-can-i-configure-it-to-work-with-recaptcha

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -68,6 +68,7 @@ module Recaptcha
       class_attribute = options.delete(:class)
       site_key = options.delete(:site_key)
       hl = options.delete(:hl).to_s
+      nonce = options.delete(:nonce)
       skip_script = (options.delete(:script) == false)
       data_attributes = {}
       [:badge, :theme, :type, :callback, :expired_callback, :size, :tabindex].each do |data_attribute|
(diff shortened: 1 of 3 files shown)


lib/recaptcha/client_helper.rb
```diff
@@ -79,7 +80,8 @@ def self.recaptcha_components(options = {})
         site_key ||= Recaptcha.configuration.site_key!
         script_url = Recaptcha.configuration.api_server_url
         script_url += "?hl=#{hl}" unless hl == ""
-        html << %(<script src="#{script_url}" async defer></script>\n) unless skip_script
+        nonce_attr = " nonce='#{nonce}'" unless nonce.empty?
```
grosser: how about skipping the `.to_s` above and then doing `if nonce` here ?

grosser:
looks about right!
also needs a test so I don't break it in the future :)

NARKOZ:
Sorry, I've added a test and applied the requested change.

grosser:
good job, nice and clean ❤️

grosser:
4.7.0

PR #266, opened 2018-04-06
Support data-error-callback. Document all callbacks

### LINKED ISSUE #265: Support data-error-callback
In V2, there is a [`data-error-callback`](https://developers.google.com/recaptcha/docs/invisible#render_param), the gem should accept an option key to generate an attribute for it.

YurySolovyov:
Fixes #265


grosser:
4.8.0

PR #268, opened 2018-04-17
LazyLoad through ActiveSupport.on_load

ericproulx:
Lazy load the gem through `ActiveSupport.on_load` instead of `require 'recaptcha/rails'`


grosser:
looks nice, any specific reason why this is done / any bug or issue you ran into ?

ericproulx:
No bug or issue, just a performance point of view. Instead of forcing `ActionView` and `ActionController` to load, just let the framework load it and include the files after. I think it's the best way in a Rails context.

grosser:
def a best-practice ... never had issues with the other way around though

PR #269, opened 2018-04-17
# frozen string literal: true

ericproulx:
Added the `# frozen_string_literal: true` magic comment for ruby 2.3 and higher


grosser:
FYI pushed an update to rubocop to display by default and move `All` to the top

PR #273, opened 2018-05-03
Validate model object first, helpful when you use xhr/ajax req

frontandstart:
I use xhr request to send and validate data, validate model object first - prevent handling errors when you already validate captcha request but not pass own validation errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ And, add `verify_recaptcha` logic to each form action that you've protected.
 ```Ruby
 # app/controllers/users_controller.rb
 @user = User.new(params[:user].permit(:name))
-if verify_recaptcha(model: @user) && @user.save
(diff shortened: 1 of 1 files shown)


grosser:
this is a reallly bad idea ... this would mean all users can bypass recaptcha and get signed up, also would be confusing if a user gets the captcha wrong and then sees a "email already taken" validation

grosser:
if you need it that can use something like `if user.valid? && recaptcha && user.save!(validate: false)`

frontandstart:
@grosser Yes you are right, sorry for wasting time. Should use more clear way to do this
like you text: if validate_model && validate_recap &&

grosser:
np, good to also have examples/discussions about what not to do around :)

Issue #278, opened 2018-06-23
w3c validation


ydakuka:
I get the following errors:

![https://image.ibb.co/bZPvUo/2018_06_23_14_42_48.png](https://image.ibb.co/bZPvUo/2018_06_23_14_42_48.png)

PR #279, opened 2018-06-23
Fix styles of Recaptcha::ClientHelper

ydakuka:
https://github.com/ambethia/recaptcha/issues/278

Fixed all errors except `scrolling="no"`.

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -25,9 +25,8 @@ module Recaptcha
                 <div style="width: 302px; height: 422px; position: absolute;">
                   <iframe
                     src="#{fallback_uri}"
-                    frameborder="0" scrolling="no"
-                    style="width: 302px; height:422px; border-style: none;">
-                    title="ReCAPTCHA"
+                    scrolling="no" name="ReCAPTCHA"
+                    style="width: 302px; height: 422px; border-style: none; border: 0;">
                   </iframe>
                 </div>
               </div>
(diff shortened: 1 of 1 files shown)


lib/recaptcha/client_helper.rb
```diff
@@ -25,9 +25,9 @@ def recaptcha_tags(options = {})
                 <div style="width: 302px; height: 422px; position: absolute;">
                   <iframe
                     src="#{fallback_uri}"
-                    frameborder="0" scrolling="no"
-                    style="width: 302px; height:422px; border-style: none;">
-                    title="ReCAPTCHA"
+                    scrolling="no"
+                    style="width: 302px; height: 422px; border-style: none; border: 0;">
+                    ReCAPTCHA
```
grosser: that has to stay the title ?

grosser:
would be great if we had tests for this too, but I guess that's pretty hard ... 
the title change looks wrong though, the error was that there is text in the iframe and now there is text again ...

grosser:
can you also explain how you got these errors for anyone wanting to change/reproduce this later

ydakuka:
I got these errors by analyzing the page with the recaptcha through the site https://validator.w3.org/

lib/recaptcha/client_helper.rb
ydakuka: I moved the title to the name attribute. It's ok?

grosser:
awesome!

grosser:
4.8.1

might also be fun to use https://validator.w3.org/ as an api and pass in generated html as a test

PR #288, opened 2018-08-07
Leave tabindex attribute alone for invisible_recaptcha_tags

steveh:
As `invisible_recaptcha_tags` generates a `<button>`, it should have a normal `tabindex` attribute, rather than a `data-tabindex` attribute to be picked up by the Recaptcha JS.


grosser:
awesome!

grosser:
4.11.1

PR #297, opened 2018-11-01
Add onload and render

Aaron-Lerner:
This allows you pass the `onload` and `render` options which are useful for explicit rendering. More detail on these options in [the docs](https://developers.google.com/recaptcha/docs/display#explicit_render).

### DIFF
--- a/lib/recaptcha/client_helper.rb
+++ b/lib/recaptcha/client_helper.rb
@@ -95,11 +95,11 @@ module Recaptcha
       unless Recaptcha::Verify.skip?(env)
         site_key ||= Recaptcha.configuration.site_key!
         script_url = Recaptcha.configuration.api_server_url
-        query_params = hash_to_query({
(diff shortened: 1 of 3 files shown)


lib/recaptcha/client_helper.rb
```diff
@@ -93,7 +95,12 @@ def self.recaptcha_components(options = {})
       unless Recaptcha::Verify.skip?(env)
         site_key ||= Recaptcha.configuration.site_key!
         script_url = Recaptcha.configuration.api_server_url
-        script_url += "?hl=#{hl}" unless hl == ""
+        query_params = hash_to_query(
```
grosser: should be `ClientHelper.hash_to_query` ?

grosser: ah nvm ... it's called from the class-method part :)

grosser:
v4.13.0

PR #299, opened 2018-12-01
Corrected value setting for ui input option.

deadcast2:
Hello! Just noticed a small error for the :ui => :input option. It now renders correctly with the supplied :text parameter.


grosser:
looks good, does the test need to change too ?

deadcast2:
Haha good catch. Missed that! ^_^

Updated the test for rendering the button as an "input" and all tests are now green.

PR #306, opened 2019-03-07
ALT: Use "https://recaptcha.net" instead of "https://www.google.com" to grant accessibility for worldwide

StephyCat:
Some countries cannot access https://www.google.com, using https://recaptcha.net would be helpful.


grosser:
tests need an update ... but in general this sounds like a safe/good change

PR #310, opened 2019-04-09
Add support for reCAPTCHA v3

TylerRick:
- Add new recaptcha_v3 helper method
- Add new verify_recaptcha_v3 method that makes the score available in the result object.
- Add config.site_key_v2, config.site_key_v3, etc. to allow using both v2 and v3 in the same app

Also:
- Add some optional logging with `config.logger`
- Extract Recaptcha::Verify::VerifyResponse, Recaptcha::Verify::VerifyResult
- Add :script_async, :script_defer, :external_script, :inline_script options


grosser:
I don't want to support multiple versions ... just switch things to v3 and we'll do a major release 🤷‍♂️ 
please split our any unrelated changes into separate PRs so this get's easier to review

TylerRick:
> I don't want to support multiple versions ... just switch things to v3 and we'll do a major release

I think you may have misunderstood what v3 is, sir. The naming makes it confusing, I agree, because it _sounds_ like v3 is a replacement for v2 (like v2 probably was for v1) since it's the "next version", after all, right? But from what I've read, that is not the case. They each serve different purposes and have different use cases. I'm using both in the app I'm using this on.

Have you read the [Should I use reCAPTCHA v2 or v3?](https://developers.google.com/recaptcha/docs/faq#should-i-use-recaptcha-v2-or-v3) section of their FAQ? I think it sums it up pretty well:

> reCAPTCHA v2 is not going away! We will continue to fully support and improve security and usability for v2.

> reCAPTCHA v3 is intended for power users, site owners that want more data about their traffic, and for use cases in which it is not appropriate to show a challenge to the user.

> For example, a registration page might still use reCAPTCHA v2 for a higher-friction challenge, whereas more common actions like sign-in, searches, comments, or voting might use reCAPTCHA v3. To see more details, see the reCAPTCHA v3 developer guide.

A fair bit of the work I did was to make it possible to use different reCAPTCHA services together in the same app. But I discovered that it wasn't just reCAPTCHA v2 and v3 that require you to have different keys. While I was testing out all the different variants, I found out that even the v2 checkbox and the invisible reCAPTCHA require you to have a different key for each! (Which is especially a pain, by the way, when you have different keys for dev, staging, and production—oh well.) Otherwise you get an error like "This site key is not enabled for the invisible captcha."

So even if we _weren't_ going to support v2 and v3 together (which I think we should), the changes I made are still useful because they let you have both a checkbox and an invisible captcha in the same app without having to manually pass in a `site_key` to one or other of the 2 view helpers ... which is really ugly and inconsistent to have to do, by the way. (How do you decide which one should be the key in the configuration and which should be the "override" key)? I made it more consistent and logical: if you want to use the v2 checkbox widget, set the v2 checkbox key; if you want to use the invisible widget, set the invisible key; etc.

TylerRick:
> please split our any unrelated changes into separate PRs so this get's easier to review

Sorry, I'm afraid it would be too hard to separate into separate PRs at this point. Much harder than it would be for you to review it... I've already spent way longer on this than I should have, so I'm not inclined to spend a ton more time on it.

It's actually all kind of related—the doc improvements go hand-in-hand with the refactorings needed to support v3, and so on.

TylerRick:
I had a question about how we want to make the API for the v3 verify method....

Since a v3 result is no longer simply a binary "pass" or "fail", we can no longer just return `true` or `false` as the return value. So I made it return a `Result` object in the case it actually gets back a valid result from the reCAPTCHA API. But it's still returning `false` when the response token is blank (in which case it doesn't even hit the API server), or if the hostname from verification doesn't match the expected hostname, for example. It kind of bugs me that the return value can be a boolean _or_ a `VerifyResult` object. I was wondering if you or anyone had any suggestions for how best to handle that.

One option would be to make it _always_ return a `VerifyResult`. That's the option I'm leaning towards currently unless anyone has a better suggestion.

Then you wouldn't have to do an extra check in your controller in case it was `false`, like you currently have to do...

```
if recaptcha_result && recaptcha_result.valid?
  if recaptcha_result.score > 0.5
    …
```

 (And we wouldn't need to automatically store `@recaptcha_verify_result = result` to provide access to the error codes when the return value is `false` because it that would be in the return value and the caller would already have access to it.)

and you could just confidently do this instead:

```ruby
if recaptcha_result.valid?
  if recaptcha_result.score > 0.5
    …
```
We would just return a `VerifyResult` that returns false for `valid?` anywhere that we're currently returning `false`.

The only downside I can think of is that it would make it so there's less code we can share between the v2 and v3 verify (the common code is currently in a private method `_verify_recaptcha`) and the tests for each would diverge (would have to `refute verify_recaptcha.valid?` instead of `refute verify_recaptcha`, for example), but I think it's worth it if it makes the public API better.

Thoughts?

---

Another option would be to return `nil` where we're currently returning false, so that you could at least use the safe navigation operator in your controller...

```ruby
if recaptcha_result&.valid?
  if recaptcha_result.score > 0.5
    …
```

TylerRick:
I changed `verify_recaptcha_v3` to always return a `VerifyResult`, so now you can do this without having to check for a `false` return value:

```ruby
recaptcha_result = verify_recaptcha_v3(model: @user, action: 'registration')
if recaptcha_result.valid?
  if recaptcha_result.score > 0.5
    …
```

If we were okay with a backwards incompatible change (i.e., a major release), we could change the v2 `verify` to do the same...

TylerRick:
Thinking some more about whether v2 could be removed...

I think we should definitely keep support for the v2 checkbox, though I do wonder whether the v2 invisible widget is still needed. Generally speaking, v3 feels like it could be a replacement for v2 invisible. It's not a drop-in replacement though. 

I'm not super familiar with v2 invisible, but at a glance anyway, the main difference seems to be that v2 invisible can render a button. I guess you could say that **v3 appears to be analogous to ` recaptcha_v2_invisible ui: :invisible`**. I don't really understand why we would need/want the catpcha helper to render a `<button>`/`<input type="submit"> tag when we could easily do that ourselves. What's the advantage of using that over using a separate call to a `submit_tag` helper, for example? (Then we wouldn't have to support a `text:` option for controlling the text of the button.)

There may be some other options on `invisible_recaptcha_tags` that don't have an equivalent in the v3 API as well. I don't see equivalents for `expired_callback` or `error_callback` mentioned on <https://developers.google.com/recaptcha/docs/v3>, for example. 

I do love that the v3 API is simpler and I think I personally would be fine with removing support for the v2 checkbox and naming v3 as its successor. I don't know how others would feel about that removal though... Maybe that should be considered in a separate issue/MR?

TylerRick:
I came across the official PHP client library for reCAPTCHA at https://github.com/google/recaptcha. It's interesting to see how similar and different our API is to theirs when they were both developed independently... Maybe we can learn a few things from them?

---
Their example/docs:
```
$recaptcha = new \ReCaptcha\ReCaptcha($secret);
$resp = $recaptcha->setExpectedHostname('recaptcha-demo.appspot.com')
                  ->setExpectedAction('homepage')
                  ->setScoreThreshold(0.5)
                  ->verify($gRecaptchaResponse, $remoteIp);

if ($resp->isSuccess()) {
    // Verified!
} else {
    $errors = $resp->getErrorCodes();
}
```

The following methods are available:

- `setExpectedHostname($hostname)`: ensures the hostname matches. You must do
  this if you have disabled "Domain/Package Name Validation" for your
  credentials.
- `setExpectedApkPackageName($apkPackageName)`: if you're verifying a response
  from an Android app. Again, you must do this if you have disabled
  "Domain/Package Name Validation" for your credentials.
- `setExpectedAction($action)`: ensures the action matches for the v3 API.
- `setScoreThreshold($threshold)`: set a score theshold for responses from the
  v3 API
- `setChallengeTimeout($timeoutSeconds)`: set a timeout between the user passing
  the reCAPTCHA and your server processing it.

Each of the `set`\*`()` methods return the `ReCaptcha` instance so you can chain
them together.

---

Similarities:
1. Their `verify` method optionally [verifies](https://github.com/google/recaptcha/blob/master/src/ReCaptcha/ReCaptcha.php#L168) that hostname and action match, same as we do...
```
        if (isset($this->action) && strcasecmp($this->action, $initialResponse->getAction()) !== 0) {
            $validationErrors[] = self::E_ACTION_MISMATCH;
        }
```
2. They [merge](https://github.com/google/recaptcha/blob/master/src/ReCaptcha/ReCaptcha.php#L190) the error codes from reCAPTCHA response and validation errors (from a mismatched hostname or action), together into a single errors array, same as we do
```
            array_merge($initialResponse->getErrorCodes(), $validationErrors),
```

Differences:
1. They call `isSuccess()` what I was calling `valid?`. I think I do like `success?` better so changing to that (left `valid?` as an alias in case some people prefer that).
2.  We ~don't~ didn't have any equivalent to their `setScoreThreshold($threshold)`.  Added `score_threshold` option.
3.  We don't have any equivalent to their `setChallengeTimeout($timeoutSeconds)`. We should probably add that. There must be some kind of default timeout already though, because there is a `timeout-or-duplicate` error code. (I've only triggered that error by trying to verify the same token twice. I wonder how to trigger a timeout?)
4. They have a "fluent" interface so you can chain options like `setScoreThreshold(0.5)->verify()`. That's kind of neet but I haven't seen that much in Ruby APIs and doesn't seem to have much benefit over just passing option directly to `verify`.
5. They conflate the initial JSON response and the result of validations into a single [Response](https://github.com/google/recaptcha/blob/master/src/ReCaptcha/Response.php) object. Personally, I think I prefer to have the response from the API be a distinct class/type from the result of applying app-specific validations and using composition (`result` has a `response` that it delegates to as needed).

bobziuchkovski:
@TylerRick It looks like you beat me to the punch here!  Let me elaborate...

We recently introduced the `recaptcha` gem in our environment to thwart credential harvesting attacks.  We're using the v2 API in checkbox mode on our login page.  Real users click the checkbox and solve the prompt if necessary.  Bots smack into the captcha and fail to sign-in, regardless of whether they have valid credentials or not.

This is all fine and good, but it's not a great user experience.  We've decided to use the v3 API **in addition to** the v2 API so that we display and validate the v2 checkbox only if Google's v3 API indicates the request is likely coming from a bot (score > some threshold).

Hence we were going to pursue adding v3 support to this gem and ensure the two could be used side-by-side, just as you have.  So thanks for your work!  This is likely going to save me quite a bit of time on my end.

That said, I can understand a bit where @grosser is coming from on minimizing the changes here.

@grosser / @TylerRick Is there anything I can do to help get the v3 support added (and working side-by-side with v2)?  I'm happy to help carve up the PR changes to minimize either unrelated changes and/or try to ensure backwards compatibility.  I'd love to hear @grosser 's thoughts on supporting v2+v3 in general and what he'd like to see in the API.

TylerRick:
Thanks, @bobziuchkovski. That seems like a great use case: transparently using v3 and only showing a v2 checkbox if score is < some threshold. I haven't tried that method myself (I just use v3 on some pages and v2 on others currently) but it should work without problem.

@bobziuchkovski Could you try integrating v2 and v3 in your app in the way you described and see if you run into any problems — either in terms of missing functionality, or ways the API could be improved, or ways the documentation could be clearer or more helpful? (If you make any improvements, I would be happy to merge/cherry-pick them into this branch if you push up your commits.)

As far as I'm concerned, this pull request is ready to merge and should make a good starting point. This doesn't have to be the final, end-all v3 API. More follow-up commits/fixes/improvements can always be added in another PR! :smiley:

I'm already using this branch in another app. It would be very helpful to have someone else try it out and confirm it's working though! :+1: 

You can try out this branch by either adding this to your Gemfile:
```
gem 'recaptcha', github: 'TylerRick/recaptcha', branch: 'v3'
```
or (if you want a more stable tag ref, that won't change underneath you if more commits get added/amended in this branch):
```
gem 'recaptcha', github: 'TylerRick/recaptcha', tag: 'v3-r10'
```

You can refer to the documentation here: https://github.com/TylerRick/recaptcha/tree/v3
and here: https://www.rubydoc.info/github/TylerRick/recaptcha/acb4d23

Maybe something like this would work?
```ruby
result = verify_recaptcha_v3(action: 'login', score_threshold: 0.5)
if result.success?
  # Perform action
else
  if result.score_below_threshold?
    @show_checkbox_recaptcha = true
  end
  render 'new'
end
```

TylerRick:
Forgot that the controller would also have to verify the token from the checkbox captcha if it had fallen back to showing a v2 checkbox challenge. Here's a more complete example that should handle that (haven't tested it though). Also added to the Readme:

```erb
  …
  <% if @show_checkbox_recaptcha %>
    <%= recaptcha_v2_checkbox %>
  <% else %>
    <%= recaptcha_v3(action: 'login') %>
  <% end %>
  …
```

```ruby
# app/controllers/sessions_controller.rb
def create
  result = verify_recaptcha_v3(action: 'login', minimum_score: 0.5)
  checkbox_result = verify_recaptcha_v2_checkbox unless result.success?
  if result.success? || checkbox_result.success?
    # Perform action
  else
    if result.score_below_threshold?
      @show_checkbox_recaptcha = true
    end
    render 'new'
  end
end
```

TylerRick:
So far I've been fairly conservative and tried hard to keep everything backwards-compatible. If we're not as worried about backwards compatibility, I would probably make the following additional changes:

- remove `recaptcha_v2_invisible`, rename `recaptcha_v2_checkbox`, and only keep `recaptcha_checkbox` from the v2 API
- change `verify_recaptcha_checkbox` to return a `result` object that can be inspected for errors (like `verify_recaptcha_v3` does) instead of a boolean
- move all of the detailed API reference to https://www.rubydoc.info so it's all in one place, and just have getting started instructions and some examples in the Readme

grosser:
bit swamped atm, will take a look soon

grosser:
First off, thx for the PR!

 1. having a "lighhouse" PR like that, that explores all the consequences/where we should be going is great
 1. comparing and staying in sync with official libraries sounds like a great idea and would make a great major bump or just PR on it's own
 1. This PR is to big for me to understand all the changes/implications
 1. I'd prefer to break the PR up and merge piecemeal ... ideally without breaking stuff at first and then a major bump if necessary (as small as makes sense, I can review/merge that super quickly) ... also shows all the details and allows discussion on smaller details instead of just rubber-stamping the whole PR
 1. I'd prefer returning true/false from verify, the library can just consider <0.5 to be false (can change via env var)
 1. I prefer not to rely/use rubydoc.info because it does not take the users to the source and so discourages reading the source/making PRs, also does not work well when looking at older versions of the code / local checkout ... maybe link from the readme into the code to not duplicate docs and keep the docs there or so
 1. atm Zendesk still uses recaptcha, so I don't want to break it/I have some 2nd hand knowledge, but I'm not actively using it, I could also add you guys as maintainers if you want to maintain the library ... can always ping me for help ... /cc @kintner if you know anyone who wants to maintain this library

jalkoby:
just updated my project to recaptcha v3 with recaptcha gem 0.14.0 - there is no need in so many breaking changes. I will try to do a different pr with v3 support.

grosser:
thx!

TylerRick:
Sorry, I would have done it the other way if I had known your preference. I thought these changes would be seen as obvious improvements and quickly merged.

I understand where you're coming from re. smaller PRs, though from a contributor's perspective, understand that it's more work to create, track, and follow through on a bunch of small PRs. And if you want to actually use/benefit from all of your changes immediately (which I did want to do), you'd have to maintain **both** a set of isolated branches each with only a single change, **and** an alternate integration/master branch with all of them merged together. And if you wanted to make use of one refactoring in another branch, you couldn't do it without introducing inter-branch dependencies. So time is wasted waiting for the merge of one branch, then rewriting and rebasing the others. After making one change/fix, it just feels silly to reset back to master (temporarily discarding the changes you just made and pretending they don't exist) to start your next change from `master` instead of building on what you just wrote. I guess that's a luxury only maintainers get to enjoy. Anyway, that's why I did it this way, for what it's worth, though I realize my shortcut didn't save any time this time. :)

(I was actually being _fairly_ conservative; there are way more things I _would_ have refactored if this were my own project and I weren't trying to limit the scope of things. But I guess it's all relative. :) )

TylerRick:
Not sure what breaking changes you are referring to, @jalkoby . I tried to keep everything backwards compatible and all the tests still passing. It could be done with fewer refactorings, certainly, but even you couldn't resist refactoring, I see (#314). ;-)

TylerRick:
Okay, I made a new, more minimal PR for this, #317, based on top of @jalkoby's excellent recent refactorings.

I plan to follow up with a PR that allows you to access the full details about the response, including the score and any error codes that were returned. It seems crazy to leave developers totally in the dark about which of many possible validations or errors caused it to return `false`.

After that, I may or may not follow up with PRs for the other things I think would be good to have:

- The name of the existing helper `recaptcha_tags` could be improved. I would suggest `recaptcha_checkbox` to make it clear _which_ of the 3 widgets it is rendering.

- Support using v2 and v3 in same app:
  - support configuring multiple keys, and provide different helper for each widget so it can use the right key without having to pass it in to the view helper (yuck)
  - add demo example with the fallback to v2 checkbox example from Readme

- Add optional logging

grosser:
❤️ let me know if you are happy with everything and want the major bump to go out

TylerRick:
>  let me know if you are happy with everything and want the major bump to go out

Well, I do really want to see support for easy v2 and v3 in same app and ability to inspect the recaptcha result before I'm happy with it.

I could try to do that without breaking changes but I'm not sure on the best way to make the result available, especially if you want `verify_recaptcha` to only return true/false...

How about making a minor version (4.15) release with v3 support first for those who need v3 now, and then we can be free to start making a few incompatible changes in `master` in preparation for a 5.0 release?

> 5. I'd prefer returning true/false from verify, the library can just consider <0.5 to be false (can change via env var)

Can we provide methods for both ways—one that just returns true/false and one that returns an object that you can query if you require more details? And if so, what to call those 2 different methods?

How about using `recaptcha_valid?` (or `valid_recaptcha?`) if you just want true/false (so that it works similarly to `valid?` from other familiar APIs) and  `verify_recaptcha` if you want a result object?

Of course, `verify_recaptcha` is already returning true/false currently so that would require a major version to change that.

Another option would be to add a new `recaptcha_result` method that returns a result object and 
leave `verify_recaptcha` as boolean (but refactor it to reuse `recaptcha_result` and return `result.success?`). The reason I don't love this option though is because `recaptcha_result` method isn't a very strong name — it has no verb in it so it's not clear that it actually _performs_ the verification (has side effects); it might sound like it's just something you call _after_ you've done the verification to retrieve the result object that's stored in an instance variable.

We could seek another verb... like `perform_recaptcha` or `perform_recaptcha_verify`, but neither is great, because it doesn't match the verb used by the upstream API and docs, which is "verify", so adding alternate verbs that also do the verify would just be confusing. `verify_recaptcha` is really the only good name candidate I can think of.

Therefore, my recommendation is to change `verify_recaptcha` to return a query-able result object and either just make people call `.success?` on that result (consistent with how the official PHP client [does it](https://github.com/google/recaptcha#usage))
or add a `valid_recaptcha?` if people really want to save 10 key strokes and get a boolean with a single method call.


---

> the library can just consider <0.5 to be false (can change via env var)

I'd recommend not picking an arbitrary default score threshold. It's so easy to just pass it in, `verify_recaptcha(minimum_score: 0.5)`, and I believe there's a valid use case for just passively verifying recaptcha tokens _without_ setting any minimum and just storing that score for later use/reference.

I think it would be the most consistent and least surprising behavior if we only validate what the user asks us to validate; that is, only make score a factor in determining success _if_ a score is passed in. 

This is consistent with how not passing in any expected `hostname` makes the `hostname_valid?` check return true. It also appears to be how the official client [works](https://github.com/google/recaptcha#usage): you call `setScoreThreshold(0.5)` if you want it to require a certain score, and you don't call it if you don't want to add that requirement.

grosser:
- so do a minor release now ?
 - I'd prefer `recaptcha_result` since I don't expect most ppl to read the changelog and suddenly having all `verify_recaptcha` calls passing is a big security hole :D
 - use a 0.5 or whatever if we need to pass a value ... or require passing a value for v3

jalkoby:
Guys let's wait till the end of the week

TylerRick:
> Guys let's wait till the end of the week

Doesn't matter to me, though I don't know what downside there would be to doing a minor release now and then another release whenever there's more to release. Releases are cheap and easy ;)

TylerRick:
> * I'd prefer `recaptcha_result` since I don't expect most ppl to read the changelog and suddenly having all `verify_recaptcha` calls passing is a big security hole :D

Fair enough, I guess security risks might trump other concerns ... :)

ulisescarreonalvarez:
If you show recaptcha_v2_checkbox on the mismatch of the score how set the keys for that v2 recaptcha.

On the console shows  `undefined method `site_key_v2=' `

TylerRick:
@ulisescarreonalvarez,

> If you show recaptcha_v2_checkbox on the mismatch of the score how set the keys for that v2 recaptcha.

You can pass `site_key` to `recaptcha_v3` and pass `secret_key` to `verify_recaptcha`. 

> On the console shows `undefined method `site_key_v2=' `

The ability to configure this library with multiple keys  (`config.site_key_v2_checkbox =`, `config.site_key_v3 =`, ...) would be much nicer than manually trying to pass the correct `site_key` and `secret_key` to all your calls, wouldn't it? 

This pull request would have added that ability. Unfortunately, the maintainer decided not to merge it. You could try submitting a new PR with that functionality if you want...

grosser:
yeah global config for that sounds good, should definitely not have to pass
them inline
sorry if I got that wrong in the PR review :(

nicholasshirley:
Wow, great work @TylerRick and nice discussion. After doing some testing I agree with the discussion in this thread, v2 & v3 need to live together. Despite the unfortunate naming they are actually different strategies, not different versions, and are likely to be deployed simultaneously.

@grosser can you give some direction on how you want the API to develop in light of this? IMO Even if separate version keys (and the invis key) are accomodated, it seems like there is going to be a lot of confusion calling methods without their explicit version number unless there is a breaking changes to the method names.

grosser:
either keep the old methods working with an alias or break them obviously
by using completely new methods
I'd kinda have v2 / v3 things in the method names because that seems rather
dirty, but if that's the simplest way then go for it ...

PR #311, opened 2019-05-08
Refactor code by splitting interface and implementation

jalkoby:
Hi, I've made some refactoring before adding v3 support - as all code is put together. so the 1st step will be separate adapters & implementation


grosser:
- constant cleanup looks good
 - would prefer `ViewMethods/ControllerMethods` ... move the static methods that are reused to just `Recaptcha` so we can delete the Verify/ClientHelper
 - do we need to make all these methods static <-> keeping them on the instance makes them overridable ... at least make them static on the `View ` so we don't have view logic in 2 different files if possible
 - needs a changelog entry to say X changed to Y so it's easy to upgrade
 - would be easier to digest if this was 3 PRs with `bump bundler` + `constant cleanup` + `rename stuff`

jalkoby:
Splitting into 3 and change log update makes sense. Will do tomorrow- eu zone)

jalkoby:
btw here is a post by Yehuda Katz - https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/ maybe it's better to ignore gemfile.lock as they are related to dev env only

jalkoby:
closed in favour of 3 separated prs #312 #313 #314

PR #313, opened 2019-05-09
Cleanup constants

jalkoby:
The 2nd part of #311 - cleanup constants (after #312)


.gitignore
```diff
@@ -6,3 +6,5 @@ pkg
 
 .gems
 .rbenv-gemsets
+
+Gemfile.lock
```
grosser: I prefer keeping it to make contribution more stable https://grosser.it/2015/08/14/check-in-your-gemfile-lock/

.travis.yml
```diff
@@ -15,4 +15,6 @@ matrix:
 script: bundle exec rake $TASK
 branches:
   only: master
-before_install: ruby -e "File.write('Gemfile.lock', File.read('Gemfile.lock').split('BUNDLED WITH').first)"
+before_install:
+  - gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true
+  - gem install bundler -v '< 2'
```
grosser: faster / simpler:
`before_install: ruby -e "File.write('Gemfile.lock', File.read('Gemfile.lock').split('BUNDLED WITH').first)"`

demo/rails/Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
 gem 'dotenv-rails', groups: [:development, :test]
-gem 'rails', '4.2.5'
+gem 'rails', '~> 4.2.11.1'
```
grosser: any v4 should do, so this is better/less locked down:
```suggestion
gem 'rails', '~> 4.2'
```

recaptcha.gemspec
```diff
@@ -23,5 +23,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency "pry-byebug"
   s.add_development_dependency "bump"
   s.add_development_dependency "webmock"
-  s.add_development_dependency "rubocop"
+  s.add_development_dependency "rubocop", ">= 0.68.0"
```
grosser: no need for that if we have a Gemfile.lock :)

.travis.yml
jalkoby: actually in #314 I've removed it - all builds works fine

PR #314, opened 2019-05-09
Refactor internals

jalkoby:
The 3rd part of #311 - refactoring and extraction of adapters (after #312 #313). I did not understard your comment about override methods as there were no such info in the docs


README.md
```diff
@@ -57,8 +57,8 @@ See [sinatra demo](/demo/sinatra) for details.
 
  - add `gem 'recaptcha'` to `Gemfile`
  - set env variables
- - `include Recaptcha::ClientHelper` where you need `recaptcha_tags`
- - `include Recaptcha::Verify` where you need `verify_recaptcha`
+ - `include Recaptcha::Adapters::ViewMethods` where you need `recaptcha_tags`
+ - `include Recaptcha::Adapters::ControllerMethods` where you need `verify_recaptcha`
```
grosser: changelog entry for this ?

TylerRick:
Nice refactorings! Definitely a step in the right direction.

I'm curious why we need/want an `Adapters` module/namespace. Couldn't we just put those `ViewMethods` and `ControllerMethods`  directly under `Recaptcha::`? The word Adapters in there doesn't seem to add anything useful and is puzzling me.

Also wondering, could we come up with a better name for `.verify_via_api_call` or `.api_verification` (formerly `.get`)? They both sound like the same thing.

jalkoby:
Hi. The idea of adapter is provide connection to the library without introducing extra methods in views/controllers

PR #320, opened 2019-06-05
Recaptcha.net for international users

jackweinbender:
Apologies for the duplicate PR, I was feeling impatient (See #306). For reference, the google documentation [can be found here](https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally):

PR #321, opened 2019-06-13
use recaptcha.net

jhanifen:


jhanifen:
Apologize, meant to commit this to our forked version.  Although this is a change that seems worth making for all users so the recaptcha works better globally per the Google FAQ listed here https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally

grosser:
hmmm this should be read from the config ?

grosser:
followup for https://github.com/ambethia/recaptcha/pull/320

demo/rails/app/views/captchas/index.html.erb
```diff
@@ -35,7 +35,7 @@
     <div id="example3"></div>
     <%= submit_tag %>
   <% end %>
-  <script src="https://www.google.com/recaptcha/api.js?onload=onloadCallback&render=explicit" async defer></script>
+  <script src="https://www.recaptcha.net/recaptcha/api.js?onload=onloadCallback&render=explicit" async defer></script>
```
grosser:
```suggestion
  <script src="<%= Recaptcha.configuration.api_server_url %>?onload=onloadCallback&render=explicit" async defer></script>
```

jhanifen: that would work great and makes a lot more sense.

grosser:
5.0.1

grosser:
derp ... this is just demo code, I yanked 5.0.1

PR #324, opened 2019-07-15
Look for keys in encrypted credentials and ENV

nicholasshirley:
Assign site key and secret with encrypted credentials if no ENV variable is
given. The encrypted secrets values need to be at the top level of the
yml file as noted in the readme.

My understanding is that (encrypted) secrets are [going to be deprecated](https://guides.rubyonrails.org/5_2_release_notes.html#credentials) eventually, but I left the language in now because I didn't know if it's going to happen in Rails 6 or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Note: Enter `localhost` or `127.0.0.1` as the domain if using in development wit
 gem "recaptcha"
 ```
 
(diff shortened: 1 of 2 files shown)


nicholasshirley:
Sorry...I can't read. I see the alternative key setup instructions now.

grosser:
np, thx for trying to make a PR anyway :D

PR #327, opened 2019-08-02
Allow addition of more translations

### LINKED ISSUE #326: In-gem translations?
I can see from https://github.com/ambethia/recaptcha#i18n-support that `recaptcha.errors.verification_failed` and `recaptcha.errors.recaptcha_unreachable` are the keys to use for localization. I was going to add these to my project which supports 34 languages, but I figured it may be better to share the work with everyone.

Is there interest to add the translations to the gem directly? I can point my translators here once it's set up.

JasonBarnabe:
Closes #326.

Adds `rails/locales` directory and reads translations it. `en.yml` included, more can be added later!

### DIFF
--- a/lib/recaptcha/railtie.rb
+++ b/lib/recaptcha/railtie.rb
@@ -18,16 +18,18 @@ module Recaptcha
       end
     end
 
-    protected
+    class << self
+      protected
 
-    def self.add(pattern)
-      files = Dir[File.join(File.dirname(__FILE__), '../..', pattern)]
-      I18n.load_path.concat(files)
-    end
+      def add(pattern)
+        files = Dir[File.join(File.dirname(__FILE__), '../..', pattern)]
+        I18n.load_path.concat(files)
+      end
 
-    def self.pattern_from(args)
-      array = Array(args || [])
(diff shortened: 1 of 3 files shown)


grosser:
a little much magic (load-paths / finding files / instance_eval etc) for a tiny feature, I was hoping we'd just add files and rails could pick em up ... let's just do https://github.com/ambethia/recaptcha/pull/328 for now wait if anyone comes up with a cleaner solution 🤷‍♂

JasonBarnabe:
This is basically copied from https://github.com/tigrish/devise-i18n, which I maintain, but I didn't write the original code. All it's doing is saying that for each locale the hosting app uses, if we have a file for that locale, then add it to the load path.

If you merge this, then it becomes automatic to add new locales - just add another file in rails/locales. It could be de-magic-ed a bit, at the cost of performance (add each locale file individually rather than as a pattern, and/or unconditionally add every locale to the load path regardless of whether it's needed) or maintainability (require us to modify railtie.rb every time a locale is added).

grosser:
ok, let's try this ... if it's already used in a often used gem then it can't be that bad :)

grosser:
5.1.0

PR #334, opened 2019-10-01
Use default if timeout is nil

maalur:
With the current implementation the default timeout is never used because a `timeout` argument is always passed to `api_verification`. This was made obvious today when reCAPTCHA was having issues and we saw network request times as long as 2 minutes.

### DIFF
--- a/lib/recaptcha.rb
+++ b/lib/recaptcha.rb
@@ -64,7 +64,7 @@ module Recaptcha
     verify_hash = { 'secret' => secret_key, 'response' => response }
     verify_hash['remoteip'] = options[:remote_ip] if options.key?(:remote_ip)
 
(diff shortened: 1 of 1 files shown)


lib/recaptcha.rb
```diff
@@ -99,7 +99,8 @@ def self.score_above_threshold?(score, minimum_score)
     end
   end
 
-  def self.api_verification(verify_hash, timeout: DEFAULT_TIMEOUT)
+  def self.api_verification(verify_hash, timeout)
```
grosser:
```suggestion
  def self.api_verification(verify_hash, timeout: nil)
```

grosser:
thx! 5.1.1

PR #344, opened 2019-12-30
Move documentation for the Recaptcha initialiser

kohrVid:
Hi. Thanks for writing/maintaining this gem as it's very useful. I personally found the README a little bit hard to follow so I hope you don't mind my raising a PR to reorder one of the sections. I think that this should also fix a typo in the docs ("Recaptcha.configure" ought to be "Recaptcha.configur**ation**" if I'm not mistaken).
Do let me know if that's ok or if there's anything that you'd like me to change first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Add `recaptcha_tags` to the forms you want to protect:
 <% end %>
 ```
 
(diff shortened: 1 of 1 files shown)


grosser:
old ordering makes sense to me ... "configure via ENV or use one of these methods <link>"

Issue #342, opened 2019-12-18
Turbolinks and Recaptcha


paulbaum:
we use this gem in 3 scenarios. One of them does not use turbolinks and it works flawless.
The other 2 pages have turbolinks enabled. 
With recaptcha v3 and fallback v2 we experienced different errors.

The first one that came up was:
**"grecaptcha not defined"** as library loading seems to not be in sync when using turbolinks.
There were some solutions for older rails versions but using 6 nothing has worked to fix the error.
We therefore removed the immediate execution of executeRecaptchaForRequest and instead passed this as a callback into the  api.js url in order to have it executed when the library has been loaded.

This fixed the error but then the definition of the async method led to a syntax error **"unexpected token function"**. Since we already used a patch we also deleted this definition since we don't use it right now. This worked for us but is not really the good way and can crash when recaptcha gets updated.

Are there any solution I missed working with rails 6? (ajax: true didn't work for us)

grosser:
idk of any fix for this, please dig deeper and post your findings here /
make a PR :)

estani:
This does seem to work, but in the console I get an error when starting from a different page, navigating to the form, then away again and then coming back to it:
```
recaptcha__de.js:279 Uncaught Error: reCAPTCHA placeholder element must be empty
```
The form does seem to work properly though. The recaptcha badge on the bottom right is not expanding anymore though, thus not displaying the legal infos.

paulbaum:
@estani seems to be another/extended issue. I'll close this for now, you can create a new one for your specific problem :)

estani:
ok!

PR #346, opened 2020-01-02
Turbolinks and recaptcha

paulbaum:
https://github.com/ambethia/recaptcha/issues/342

as I used a monkey patch for that which works for us I'm not sure if it would work for others / doesn't destroy anything.

I'm happy for feedback and I don't think this solution is the best but it works

### DIFF
--- a/lib/recaptcha/helpers.rb
+++ b/lib/recaptcha/helpers.rb
@@ -214,7 +214,7 @@ module Recaptcha
     private_class_method def self.recaptcha_v3_onload_script(site_key, action, callback, id, options = {})
       nonce = options[:nonce]
       nonce_attr = " nonce='#{nonce}'" if nonce
-
+      console.log("test")
       <<-HTML
         <script#{nonce_attr}>
           // Define function so that we can call it again later if we need to reset it
--- a/lib/recaptcha/helpers.rb
+++ b/lib/recaptcha/helpers.rb
@@ -187,7 +187,6 @@ module Recaptcha
           function #{recaptcha_v3_execute_function_name(action)}() {
             grecaptcha.ready(function() {
               grecaptcha.execute('#{site_key}', {action: '#{action}'}).then(function(token) {
-                //console.log('#{id}', token)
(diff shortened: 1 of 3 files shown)


lib/recaptcha/helpers.rb
```diff
@@ -205,6 +211,27 @@ def self.to_message(_key, default)
...
+      nonce_attr = " nonce='#{nonce}'" if nonce
+
+      <<-HTML
+        <script#{nonce_attr}>
+          // Define function so that we can call it again later if we need to reset it
+          // This executes reCAPTCHA and then calls our callback.
+          function #{recaptcha_v3_execute_function_name(action)}() {
+            grecaptcha.ready(function() {
+              grecaptcha.execute('#{site_key}', {action: '#{action}'}).then(function(token) {
+                //console.log('#{id}', token)
```
grosser: remove ?

grosser:
- can you high-level explain how this works ? (maybe add to readme)
 - add a test ?

paulbaum:
>     * can you high-level explain how this works ? (maybe add to readme)
> 
>     * add a test ?

It basically puts the name of the function in the url used for loading googles recaptcha library as an onload callback. This way we don't need to execute the function ourselves. We still need to define it though!

turbolinks attributes true switches to the function that works that way.
If you use turbolinks - the onload property should not work any more.

lib/recaptcha/helpers.rb
paulbaum: as  this gets executed after the callback we must not remove it

grosser: I meant the console.log :)

paulbaum: sure. This is still in the original code too. I'll remove it in both cases

grosser:
can you please add the turbolinks option to the readme too ?

paulbaum:
> can you please add the turbolinks option to the readme too ?

done :)

grosser:
awesome!

grosser:
5.3.0

PR #352, opened 2020-02-15
Clarify that if using v3 -> v2 fallback example

rbarreca:
Took me a bit to realize I needed two separate keys to do v3 -> v2 callback. Adding in site and secret_key to example to clarify that concept.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ threshold:
   <% if @show_checkbox_recaptcha %>
     <%= recaptcha_tags %>
   <% else %>
-    <%= recaptcha_v3(action: 'login') %>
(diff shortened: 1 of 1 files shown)


grosser:
hmm yes, it's confusing, the default setup is to use the config, so I'd rather not recommend to always use the override options.

how about any of these solutions ?
 - separate the 2 config keys and make the methods pick the right one
 - make callbacks fail fast if the format of the key is wrong (do v2 and v3 have different formats ?)

SvcDptDev:
I had the same experience as @rbarreca. It would be good to see this simple change merged in.

grosser:
yeah agree that this is better than before

Issue #353, opened 2020-03-11
Empty g-recaptcha-response parameter in request


finist:
In some cases (not always) form with `recaptcha_v3` send empty attribute `g-recaptcha-response`, and it failed recaptcha validation. It happens because recaptcha script generates empty textarea with the same name `g-recaptcha-response` and when request is sending, it replaces `g-recaptcha-response[action_name]` param.

<img width="1331" alt="recaptcha" src="https://user-images.githubusercontent.com/237128/76415054-8d54f700-63a9-11ea-8b1c-7129fdd0bffd.png">

```
Parameters: {"utf8"=>"✓", "authenticity_token"=>"QyG5eZups7QmAf04XGNkt+9pp61tCzT0M5/WMC9jmFI80shJat8/T7QCLvsxESY6/gtBpdi2n8wpCpYwGPURVA==", "g-recaptcha-response"=>"", "feedback"=>{"name"=>"test", "email"=>"test@test.com", "subject"=>"test", "content"=>"test"}, "button"=>""}
```

For fast fix add to `recaptcha_v3` parameter `name`:
```
recaptcha_v3(name: 'g-recaptcha-response-contact', action: 'contact', inline_script: false, onload: 'executeReCaptcha')
```
don't use `[]` for name, it make attribute included, and will be replaced by empty field

and add `response` parameter to `verify_recaptcha`:
```
verify_recaptcha(response: params['g-recaptcha-response-contact'], action: 'contact')
```

Preen:
Saviour!!!

grosser:
can this be a PR that fixes the code or makes it fails in a more obvious way ?

slehmann36:
I ran into this issue today as well. @finist's suggested workaround does seem to work, however, why is it necessary? This has been working in production for two months now, yet today, it became an issue - nothing's changed.

Everything was set up using just `verify_recaptcha(action: "sign_in")` in the controller, and `recaptcha_v3(action: "sign_in")` on the view, as per documentation. 

Any ideas why this would stop working? Has Google made a change to their reCAPTCHA V3 API perhaps?

eduanttunes:
It stopped working here too.
`captcha = verify_recaptcha(action: 'login', minimum_score: 0.5)`
It always return false, and an empty response from Google.
What should I do?

eduanttunes:
Life saver! @finist

esale:
In my case, I had to leave `inline_script: true` for this great workaround to apply.

[...]

PR #356, opened 2020-03-17
fix. Not working params name.

fooramu:
Yesterday updated parameter 

`params['g-recaptcha-response-data']
`
is not working  in「reCAPTCHA v2」

Real response params name is 'g-recaptcha-response'

### DIFF
--- a/lib/recaptcha/adapters/controller_methods.rb
+++ b/lib/recaptcha/adapters/controller_methods.rb
@@ -75,7 +75,7 @@ module Recaptcha
       # the key.
       # @return [String] A response token if one was passed in the params; otherwise, `''`
       def recaptcha_response_token(action = nil)
-        response_param = params['g-recaptcha-response-data']
+        response_param = params['g-recaptcha-response']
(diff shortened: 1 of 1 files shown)


grosser:
CI failed ... how about changing the input to be -data instead so everything is consistent ?
(or do we not control that ?)

fooramu:
Does that mean that the following changes are not particularly mistakes?

`g-recaptcha-response ---> g-recaptcha-response-data`

https://github.com/ambethia/recaptcha/commit/a07e0caee48ad11c05971eecbb728c5039aac033
I don't understand what perpose this commit.

finist:
@fooramu #353 this issue was purpose of this commit. If change attribute name to `g-recaptcha-response` it broke recaptcha_v3

Zhao-Andy:
Hi, also experienced this issue using v2 `recaptcha_tags`. For a short term solution, I did this:
```ruby
verify_recaptcha(secret_key: your_app_secret, response: params["g-recaptcha-response"])
```

grosser:
5.4.1

PR #357, opened 2020-03-17
fix verify_recaptcha for v2

finist:
For support v2 and v3 we need to check `g-recaptcha-response-data` and  `g-recaptcha-response` parameters


grosser:
5.4.1

Issue #343, opened 2019-12-18
Getting the v3 humanity score


esale:
Hi,

I'd like to get the humanity score reported by Google. In you doc says

> You need to verify this token with verify_recaptcha in your controller in order to get the score.

but I can't get the score, only a boolean.

Is there any way to get the actual number? There is another gem, new_google_recaptcha that can report the score to the controller.

Regards and thanks in advance.

grosser:
need to call api_verification manually like done here https://github.com/ambethia/recaptcha/blob/43c1a5dd460c6ee2abf4dd6d59775a4f79cd0fd7/lib/recaptcha.rb#L67
PR to optionally return the full response instead of true would also be welcome since that was asked for before

esale:
Hi,

I didn't do a PR, but I think maybe [the code in this gist](https://gist.github.com/esale/c4d08377d5df5ce6b2540358946e16a3) could be added to your project.

Is it helpful?

I used verify_recaptcha as skeleton. In my project, I'm currently using what you suggested as:

`Recaptcha.api_verification({secret: Recaptcha.configuration.secret_key, response: recaptcha_response_token("contact")})`

and it works fine.

Thanks and regards.

grosser:
looks about right PR would be nice :)

Intrepidd:
I have started working on it in #358, would love your opinion on it so we can make it happen !

grosser:
5.5.0

PR #358, opened 2020-03-30
Expose recaptcha reply (and score) to the controller

Intrepidd:
As mentioned in #343

It would be very helpful to be able to get the recaptcha v3 score, for logging purposes for example, or to have different responses given the score (more than just a binary yes/no)

This PR is WIP, I just want to validate the implementation, as I'm adding some public methods and tweaking a bit the `verify_via_api_call` by returning a tuple which is a bit weird.

If you are ok with the implem, I will add tests and update the doc.
Otherwise, if you have suggestions I'm all ears

Thanks

### DIFF
--- a/lib/recaptcha.rb
+++ b/lib/recaptcha.rb
@@ -65,10 +65,16 @@ module Recaptcha
     verify_hash['remoteip'] = options[:remote_ip] if options.key?(:remote_ip)
 
     reply = api_verification(verify_hash, timeout: options[:timeout])
-    reply['success'].to_s == 'true' &&
+    success = reply['success'].to_s == 'true' &&
       hostname_valid?(reply['hostname'], options[:hostname]) &&
       action_valid?(reply['action'], options[:action]) &&
       score_above_threshold?(reply['score'], options[:minimum_score])
+
+    if options[:with_reply] == true
+      return success, reply
+    else
+      return success
+    end
   end
 
   def self.hostname_valid?(hostname, validation)
--- a/lib/recaptcha/adapters/controller_methods.rb
+++ b/lib/recaptcha/adapters/controller_methods.rb
@@ -24,7 +24,9 @@ module Recaptcha
               options = options.merge(remote_ip: remoteip.to_s) if remoteip
             end
 
-            Recaptcha.verify_via_api_call(recaptcha_response, options)
+            success, @_recaptcha_reply =
+              Recaptcha.verify_via_api_call(recaptcha_response, options.merge(with_reply: true))
+            success
           end
 
           if verified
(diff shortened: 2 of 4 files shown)


lib/recaptcha.rb
```diff
@@ -65,10 +65,12 @@ def self.verify_via_api_call(response, options)
     verify_hash['remoteip'] = options[:remote_ip] if options.key?(:remote_ip)
 
     reply = api_verification(verify_hash, timeout: options[:timeout])
-    reply['success'].to_s == 'true' &&
+    success = reply['success'].to_s == 'true' &&
       hostname_valid?(reply['hostname'], options[:hostname]) &&
       action_valid?(reply['action'], options[:action]) &&
       score_above_threshold?(reply['score'], options[:minimum_score])
+
+    return success, reply
```
esale: `return [success, reply]`

grosser: how about not breaking anyones existing code ? :)
```suggestion
    if options[:reply]
      [success, reply]
    else
      success
    end
```

esale: Developer's wisdom :)

Intrepidd: I find my version simpler to read, why do you suggest this more verbose approach ?

Intrepidd: Nice thanks. How about something a little more explicit like `include_reply` or `with_reply` ?

esale: You must return an array. But in any case, @grosser already gave us a better solution.

Intrepidd: `return success, reply` is exactly the same as `return [success,reply]` :) 

Such as calling `foo(a: 42)` is the same as calling `foo({a: 42})`, just a style difference / preference

grosser: either is fine 👍

Intrepidd: Fab, if the rest is ok with you I'll work on adding tests and doc tomorrow

lib/recaptcha/adapters/controller_methods.rb
```diff
@@ -24,7 +24,8 @@ def verify_recaptcha(options = {})
               options = options.merge(remote_ip: remoteip.to_s) if remoteip
             end
 
-            Recaptcha.verify_via_api_call(recaptcha_response, options)
+            success, @_recaptcha_reply = Recaptcha.verify_via_api_call(recaptcha_response, options)
```
grosser: this will have to pass in `with_reply: true` or something like that

grosser: ah nvm ... `a, b = true` works out fine ... so the user can decide if they want the reply

lib/recaptcha.rb
grosser: yeah, just needs test + docs :)

Intrepidd:
Alright @grosser, let me know if this is missing anything !

grosser:
5.5.0

Intrepidd:
Awesome thanks !

PR #361, opened 2020-05-30
Fix execute on submit demo example

ncs1:
-Async function version is being generated only when inline_script is true.

-Fix element id
ID generated are structured like "g-recaptcha-response-data-#{action}"


grosser:
thx!

PR #366, opened 2020-10-01
Allow multiple invisible recaptchas on a single page by setting custom selector

timobleeker:
I ran into an issue where I had two forms with `invisible_recaptcha_tags` on a single page. The main form is a contact-us form, while a second newsletter-subscribe form is always present in the footer.

When submitting either form, it would then always submit the form for the _first_ button (`var form = closestForm(eles[0]);`) it could find.

This PR is an attempt to add functionality for passing an optional `selector` prop which lets you have multiple recaptcha-protected forms on a single page, as long as you uniquely name your selector props.

Curious to what you think of this. I realise it's a bit of an edge-case.

### DIFF
--- a/lib/recaptcha/helpers.rb
+++ b/lib/recaptcha/helpers.rb
@@ -139,7 +139,7 @@ module Recaptcha
       skip_script = (options.delete(:script) == false) || (options.delete(:external_script) == false)
       ui = options.delete(:ui)
 
-      data_attribute_keys = [:badge, :theme, :type, :callback, :expired_callback, :error_callback, :size, :selector]
+      data_attribute_keys = [:badge, :theme, :type, :callback, :expired_callback, :error_callback, :size]
       data_attribute_keys << :tabindex unless ui == :button
       data_attributes = {}
       data_attribute_keys.each do |data_attribute|
@@ -271,8 +271,7 @@ module Recaptcha
     private_class_method def self.default_callback(options = {})
       nonce = options[:nonce]
       nonce_attr = " nonce='#{nonce}'" if nonce
-      selector_tag = options[:selector]
-      selector_attr = options[:selector] ? "[data-selector='#{selector_tag}']" : '.g-recaptcha'
+      selector_attr = options[:id] ? "##{options[:id]}" : ".g-recaptcha"
 
       <<-HTML
         <script#{nonce_attr}>
--- a/lib/recaptcha/helpers.rb
+++ b/lib/recaptcha/helpers.rb
@@ -139,7 +139,7 @@ module Recaptcha
       skip_script = (options.delete(:script) == false) || (options.delete(:external_script) == false)
       ui = options.delete(:ui)
 
-      data_attribute_keys = [:badge, :theme, :type, :callback, :expired_callback, :error_callback, :size]
+      data_attribute_keys = [:badge, :theme, :type, :callback, :expired_callback, :error_callback, :size, :selector]
       data_attribute_keys << :tabindex unless ui == :button
       data_attributes = {}
       data_attribute_keys.each do |data_attribute|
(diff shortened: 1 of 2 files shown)


timobleeker:
Not sure who's the maintainer, but tagging @grosser as I notice you've most recently merged in this repository.

Any chance you'd be interested to merge this work in, or do you maybe have a different approach you'd like me to try?

grosser:
I'd prefer something more automated were it finds the form around itself or something like that :(

lib/recaptcha/helpers.rb
```diff
@@ -271,6 +271,8 @@ def self.recaptcha_v3_default_callback_name(action)
     private_class_method def self.default_callback(options = {})
       nonce = options[:nonce]
       nonce_attr = " nonce='#{nonce}'" if nonce
+      selector_tag = options[:selector]
+      selector_attr = options[:selector] ? "[data-selector='#{selector_tag}']" : '.g-recaptcha'
```
grosser, line 275: I'd have expected something like this:
```suggestion
      selector_attr = options[:selector] || '.g-recaptcha'
```

timobleeker:
@grosser Yeah I agree something automated would be better, but I wasn't really sure how to approach that to be honest. I also didn't want to interfere with existing implementations and this was the simplest way.

If you have an idea on how to automate it more I'm happy to work on it.

grosser:
is there some js method that gives us closest `g-recaptcha` to the clicked element ?

lib/recaptcha/helpers.rb
timobleeker: I don't think that would work, would it? For `data-selector="my-selector"`, `options[:selector]` would return `my-selector`. As far as I know, to select by a `data-` attribute you need the whole thing?

timobleeker:
@grosser Well there is [.closest()](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) (polyfill needed for IE), but I'm not sure how to get the element that's clicked on, as the callback function only gets called with the `g-recaptcha-response` token, not with the click event.

🤔 I might be missing something though.

grosser:
yeah, if we don't have a `this` then we need the selector hack :(

lib/recaptcha/helpers.rb
timobleeker: I could inline it though:
``` ruby
selector_attr = options[:selector] ? "[data-selector='#{options[:selector]}']" : '.g-recaptcha'
```

grosser: let the users pass a query-selector, that seems to be the most obvious api to me ... they can still send in `[data-selector='whatever']` ?

timobleeker: Not a 100% sure if I understand. My intention was to allow:
``` ruby
<%= invisible_recaptcha_tags selector: "my-selector" %>
```
It then gets transformed to `data-selector='my-selector'` by the `components` method. Hence the code in `.default_callback` is using `data-selector=`.

So are you suggesting to do something like this?
``` ruby
<%= invisible_recaptcha_tags selector: ".my-class", class: ".my-class" %>
# or
<%= invisible_recaptcha_tags selector: "[data-selector='my-selector']", "data-selector": ".my-selector" %>
```

grosser: ah I think I get it now, the selector attr goes into the tag anyway
how about using `id` for that, it's supposed to be unique and would make a fast selector too ?

timobleeker: Ah yeah using the id will probably be better! I'll make the change in this PR in the next few days.

lib/recaptcha/helpers.rb
```diff
@@ -271,6 +271,7 @@ def self.recaptcha_v3_default_callback_name(action)
     private_class_method def self.default_callback(options = {})
       nonce = options[:nonce]
       nonce_attr = " nonce='#{nonce}'" if nonce
+      selector_attr = options[:id] ? "##{options[:id]}" : ".g-recaptcha"
```
timobleeker, line 274: @grosser I'm now using the `id` directly. 

Alternatively we could decouple it so an id would only be used if you also pass in a `selector` matching the id, but I guess that would just over-complicate things.

grosser:
thx!
5.6.0

PR #371, opened 2020-10-23
Add Truffleruby head to CI

gogainda:


grosser (changes requested):
needs allow failure since head will always be brittle

grosser:
it works :D

PR #379, opened 2021-02-04
Adding a table of contents in README 

Vin0uz:
Also, adding a small note about which version is used in the rails setup example.


grosser:
usually not a big fan of TOCs, but yeah the readme is kinda messy :/

Vin0uz:
> usually not a big fan of TOCs, but yeah the readme is kinda messy :/

Not really a big fan neither tbh, but I was getting really lost in the readme >< lots of infos, very dense.

PR #394, opened 2021-06-18
Grecaptcha not defined patch

arslanmurtaza:
Why and what is being done.
Grecaptcha not defined error was being raised occasionally so, added window.grecaptcha check to execute recaptcha ready method only if grecaptcha is present.
## Pre-Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Next
 
+## 5.8.1
+* added `window.grecaptcha` check to execute recaptcha ready method only if grecaptcha is present
+
 ## 5.8.0
 * Add support for the enterprise API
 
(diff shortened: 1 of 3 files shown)


grosser:
why do you think it is not defined ?
... this will still not make recaptcha work, are you ok with that ?

arslanmurtaza:
@grosser I believe it is not defined due to network latency.
The recaptcha endpoint might take too much time to load, or if it might not load for some reason.

In such cases grecaptcha isn't defined.

grosser:
if it's not defined it won't work and give no indication on why it did not work, so how is this better ? / what is the problem with it currently showing an error (end users won't see it afaik) ?

arslanmurtaza:
Got your point, can you please let me know how can we make sure that grecaptcha is always loaded first before calling grecaptcha.ready?
We have rollbar setup on our project, and it raises error there occasionally.

grosser:
check for defined in a loop ?
... might be some bots triggering it, so not sure how it happens

Issue #387, opened 2021-03-03
TypeError with oddly shaped "g-recaptcha-response" params


docstun:
Hi there,

I investigated some server errors and found exceptions coming from Recaptcha `TypeError: no implicit conversion from nil to integer`

    gems/recaptcha-5.3.0/lib/recaptcha/adapters/controller_methods.rb:80:in `recaptcha_response_token': no implicit conversion from nil to integer (TypeError)
    from gems/recaptcha-5.3.0/lib/recaptcha/adapters/controller_methods.rb:16:in `verify_recaptcha'

The `g-recaptcha-response` param looks suspicious and seems to be submitted by a spammer:

    ...
    "g-recaptcha-response": [
      "\n                "
    ],
    ...

I think it would be great if Recaptcha would handle this and sanitize the param if it's an (empty) array with Recaptcha v2.

I am using Recaptcha v2 with gem version 5.3.0 but I believe this is still an issue in 5.7.0

grosser:
PR welcome, should be fairly simple to add this

PR #395, opened 2021-07-08
Handle invalid params in ControllerMethods#recaptcha_response_token

AMHOL:
Fixes: #387

This gracefully handles a situation in which requests are made without a valid `g-recaptcha-response-data` or `g-recaptcha-response` parameter, usually these requests come from automated scripts or bots, but it is preferable to not raise an exception (`TypeError: no implicit conversion from nil to integer`) in this case.

## Pre-Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Next
+* Gracefully handle invalid params
 
 ## 5.8.0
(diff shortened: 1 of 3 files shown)


grosser (approved):
seems like a lot of tests for so little code, but 🤷

grosser:
5.8.1

PR #401, opened 2022-02-11
Feature: Adds api response logging.

rimian:
This adds a logger configuration option. This is helpful for debugging mysterious failures.

It requires the logger respond to :info.

```
    Recaptcha.configure do |config|
      config.site_key              = '0000000000000000000000000000000000000000'
      config.secret_key            = 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
      config.enterprise_api_key    = 'ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ'
      config.logger = Logger.new(STDOUT) # or Rails.logger
    end
```

## Pre-Merge Checklist
sorry, I haven't written this... should I put it in "next"?

### DIFF
--- a/lib/recaptcha.rb
+++ b/lib/recaptcha.rb
@@ -176,7 +176,11 @@ module Recaptcha
   end
 
   def self.logger(message)
-    Recaptcha.configuration.logger.info(message) if logger?
+    Recaptcha.configuration.logger.info(tagged_message(message)) if logger?
+  end
+
+  def self.tagged_message(message)
+    message.is_a?(Hash) ? message.merge(Recaptcha.configuration.logger_tags) : message
   end
 
   def self.logger?
--- a/lib/recaptcha.rb
+++ b/lib/recaptcha.rb
@@ -60,11 +60,15 @@ module Recaptcha
   end
 
   def self.verify_via_api_call(response, options)
-    if Recaptcha.configuration.enterprise
+    verification = if Recaptcha.configuration.enterprise
       verify_via_api_call_enterprise(response, options)
     else
       verify_via_api_call_free(response, options)
     end
+
+    logger(verification.last)
+
+    verification
   end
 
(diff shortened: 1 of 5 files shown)


lib/recaptcha.rb
```diff
@@ -60,11 +60,15 @@ def self.invalid_response?(resp)
...
 
   def self.verify_via_api_call(response, options)
-    if Recaptcha.configuration.enterprise
+    verification = if Recaptcha.configuration.enterprise
       verify_via_api_call_enterprise(response, options)
     else
       verify_via_api_call_free(response, options)
     end
+
+    logger(verification.last)
```
grosser, line 69: not a fan of the indirection and duplicated logic
prefer something like:
```
Recaptcha.configuration.logger&.info success: success, reply: reply, event: "recaptcha-response"
```

grosser: also would help to pull the `with_reply` logic into this method so we always have the reply and can log it

grosser:
not sure if that is even needed, users can request `with_reply: true` and then do their own logging ?

rimian:
> not sure if that is even needed, users can request `with_reply: true` and then do their own logging ?

Oh, I might give that a go... thanks!

lib/recaptcha.rb
rimian: I moved the method to use `reply`. It does some tagging too.

rimian:
> not sure if that is even needed, users can request `with_reply: true` and then do their own logging ?

I was finding it hard to debug our implementation of this. We are using a logger but when exceptions are raised, it gets a bit fiddly. I just wanted the gem to log responses from the API and I could just forget about it.

grosser:
does https://github.com/ambethia/recaptcha/pull/402 work ?

PR #309, opened 2019-04-01
Fail fast for long g-responses

yifeiwu:
Submitting a verify request with a long (~4000) `g-recaptcha-response` will cause the endpoint to return a 400 error instead of a 200 response with a json body. For even larger payloads, this will return a 413 error.

This causes an error in the reply parsing since we're expecting a json response. https://github.com/ambethia/recaptcha/blob/master/lib/recaptcha/verify.rb#L79

Verifying this from https://www.google.com/recaptcha/api2/demo, the actual g-recaptcha-response should only be ~335 characters long. 

### Duplication 

Make a request with 4095 characters in the `response` parameter. Note doing this this in browsers may truncate the request. 
```
[...]

PR #403, opened 2022-03-05
Configure response limit

theirix:
Introduce a configurable response limit for the response.
PR #309 introduced a pre-check for captcha response. A limit of 4000 bytes works good for reCapctha but API-compatible hCaptcha produces much larger responses. This PR allows to set a custom limit and therefore use hHaptcha.

## Pre-Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Next
 * Gracefully handle invalid params
+* allow configuring response limit
 
 ## 5.8.0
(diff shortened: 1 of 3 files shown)


grosser (approved):
can you add something to the readme about how to use hcaptcha, since this is not very obvious :D

theirix:
Sure, good idea!

grosser (approved):
❤️

PR #420, opened 2022-09-04
Fix minimum_score for enterprise API

basabin54:
Setting the minimum_score attribute to anything would cause the v3 check to fail automatically when using reCAPTCHA Enterprise because the score accessor was incorrect and would return nil every time. I simply corrected the accessor for the score attribute coming back from the API and it works again.

## Pre-Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Next
 
+## 5.12.2
+* Fix minimum score for enterprise
+
 ## 5.12.1
 * Fix Japanese locale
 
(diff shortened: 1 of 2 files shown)


grosser:
tests check for the old code ... please check the original code that added it for clues to why
alternatively we could do `score or fallback to this new code` if you think that would work

basabin54:
Sorry about that. I think it makes sense to fall back to the old accessor just in case. I've made that update

grosser:
v5.12.2

PR #425, opened 2023-04-09
Add option to ignore null elements in recaptcha_v3 view helper.

### LINKED ISSUE #424: TypeError: Cannot set properties of null at setInputWithRecaptchaResponseTokenForConnectProvider
Rails version 7
Using with Turbo.

Error in browser console
```
VM366:11 Uncaught (in promise) TypeError: Cannot set properties of null (setting 'value')
    at setInputWithRecaptchaResponseTokenForConnectProvider (<anonymous>:11:27)
    at <anonymous>:5:17
```
Code snippet where the issue is happening
```javascript
function executeRecaptchaForConnectProvider() {
            grecaptcha.ready(function() {
              grecaptcha.execute('<hidden site key>', {action: 'connect_provider'}).then(function(token) {
                setInputWithRecaptchaResponseTokenForConnectProvider('g-recaptcha-response-data-connect-provider', token)
              });
            });
          };
                    var setInputWithRecaptchaResponseTokenForConnectProvider = function(id, token) {
            var element = document.getElementById(id);
            element.value = token; // element is null
          }
```

victor-fdez:
This pull request adds an option to the recaptcha_v3 view helper called ignore_no_element. This option is added to avoid a common issue in which the element of the callback function is null after a modal is closed, causing the function to throw an error. This option is not required, and if it is not set, the view helper will continue to work as usual. This pull request was made to resolve the issue requests in the repo recaptcha. 

image

Fixes #424 

Let me know if there anything else you would like me to add.

- added test cases
- changelog message
- readme documentation
- backwards compatible changes to recaptcha_v3
-  all passing tests

### DIFF
--- a/lib/recaptcha/helpers.rb
+++ b/lib/recaptcha/helpers.rb
@@ -16,7 +16,7 @@ module Recaptcha
       options[:render] = site_key
       options[:script_async] ||= false
       options[:script_defer] ||= false
-      options[:ignore_no_element] ||= false
+      options[:ignore_no_element] = options.key?(:ignore_no_element) ? options[:ignore_no_element] : true 
       element = options.delete(:element)
       element = element == false ? false : :input
       if element == :input
--- a/lib/recaptcha/helpers.rb
+++ b/lib/recaptcha/helpers.rb
@@ -16,7 +16,7 @@ module Recaptcha
       options[:render] = site_key
       options[:script_async] ||= false
       options[:script_defer] ||= false
-      options[:ignore_no_element] = options.key?(:ignore_no_element) ? options[:ignore_no_element] : true 
+      options[:ignore_no_element] = options.key?(:ignore_no_element) ? options[:ignore_no_element] : true
       element = options.delete(:element)
       element = element == false ? false : :input
       if element == :input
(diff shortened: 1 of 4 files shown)


grosser (approved):

grosser:
@victor-fdez can we make this the default instead ?

victor-fdez:
> @victor-fdez can we make this the default instead ?

@grosser 

Yes, that is a good suggestion. I have already modified the code to set the :ignore_no_element option to default true. Thank you for your feedback and for reviewing the PR.

lib/recaptcha/helpers.rb
```diff
@@ -235,13 +237,24 @@ def self.to_message(_key, default)
...
-            var element = document.getElementById(id);
-            element.value = token;
-          }
-      HTML
+    private_class_method def self.recaptcha_v3_define_default_callback(callback, options)
+      if options[:ignore_no_element]
+        <<-HTML
+            var #{callback} = function(id, token) {
+              var element = document.getElementById(id);
+              if (element !== null) {
```
grosser, line 245: what are you thoughts on either:
- always checking and not having an option at all
- using a single code block that inlines the option
?

```suggestion
              if (element !== null || #{!options[:ignore_no_element]}) {
```

victor-fdez: I think having the option as false by default will help us catch other edge cases, but we could consider making it true by default in the future. Using a single code block would simplify the code, so I'll make the changes accordingly. Let me know if you have any further suggestions or if you'd like me to rework the changes into the PR.

grosser: all good, I'll merge once it's a single code block 👍

PR #426, opened 2023-04-17
Permit actions as symbol

AndyStabler:
This commit allows the user to pass in the expected action as a symbol.

Previously we would compare the string action with the symbol action and we'd think the action was invalid because `'homepage' != :homepage`.

Thanks! 😄 

## Pre-Merge Checklist

_(I may need some guidance on updating the CHANGELOG—does this require a version change? )_


grosser (approved):

grosser:
5.13.1

AndyStabler:
Thanks for getting that out @grosser! 🙇🏻

PR #437, opened 2024-01-15
Add key setup to v3 example in README

Tom-Boyd:
I found it a bit confusing to understand how the keys worked when initially going through the examples. The v3 keys were set per call, while the v2 keys were set using environment variables.

I've updated the example to indicate where the v2 keys come from.

## Pre-Merge Checklist


grosser (approved):
thx!

PR #440, opened 2024-01-24
Remove unnecessary id on textarea

### LINKED ISSUE #439: id on textarea causes duplicate ids if recaptcha_tags is called twice on the same page
Our automated accessibility check is giving a warning about this.
https://github.com/ambethia/recaptcha/blob/master/lib/recaptcha/helpers.rb#L77

Could that ID be derived from `options[:id]` if present?

Is there a reason that element needs an id?

jcoyne:
Why and what is being done.

This was unused and may cause accessibility concerns if there is more than one recaptcha on the page due to multiple elements with the same id

Fixes #439 

## Pre-Merge Checklist


grosser (approved):
yeah can't see it being used for anything
only gotcha might be someone using it for styling or custom js but kinda doubt that

PR #458, opened 2024-11-26
Remove inline_script attribute from markup

taylor-steve:
The 'inline_script' option is leaking out as an attribute for the input element. This is getting flagged by some a11y tools. For example, from https://validator.w3.org/nu:

`Error: Attribute inline_script not allowed on element [input](https://html.spec.whatwg.org/multipage/#the-input-element) at this point.`

I don't think it is necessary in the markup, but alternatively we could move it to a data attribute if there's a need I'm not seeing.


grosser (approved):
ah yeah that should be safe since it's duped above 👍

grosser:
5.17.1

Issue #461, opened 2025-01-07
minimum length validation for invalid_response?(resp) method in Recaptcha module


Akashkarmakar787:
Spammers can pass random value for the token param (ex: g-recaptcha-response) which can increase the API validation call to recaptcha, consequently incurring more cost. 

Most of the Spam attacks we observed attacher sending 1-10 length chars. 
It is good to have minimum length 100, validation check to avoid unnecessary API calls to recaptcha.

grosser:
yeah that sounds reasonable,
I'd assume the spammers don't target the recaptcha specifically but just fill every form field
PR welcome

Akash-Karmakar-e3082:
@grosser, you are correct. The spammers are not targeting the recaptcha specifically. They are trying to exploit XSS and CSRF attacks on the form fields. But it is increasing the recaptcha cost.

Akashkarmakar787:
@grosser , you are correct. The spammers are not targeting the recaptcha specifically. They are trying to exploit XSS and CSRF attacks on the form fields. But it is increasing the recaptcha cost.

Akashkarmakar787:
@grosser I have raised the PR: https://github.com/ambethia/recaptcha/pull/462

PR #462, opened 2025-01-13
Minimum length validation for the recaptcha token param

Akashkarmakar787:
Add minimum length validation for the recaptcha token param

Reason:
https://github.com/ambethia/recaptcha/issues/461

Spammers can pass random value for the token param (ex: g-recaptcha-response) which can increase the API validation call to recaptcha, consequently incurring more cost.

Most of the Spam attacks we observed attacher sending 1-10 length chars.
It is good to have minimum length 100, validation check to avoid unnecessary API calls to recaptcha.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,7 +588,8 @@ Recaptcha.configure do |config|
   config.secret_key = '6Lc6BAAAAAAAAKN3DRm6VA_xxxxxxxxxxxxxxxxx'
   config.verify_url = 'https://hcaptcha.com/siteverify'
   config.api_server_url = 'https://hcaptcha.com/1/api.js'
-  config.response_limit = { max: 100000, min: 100 }
+  config.response_limit = 100000
+  config.response_minimum = 100
 end
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -588,7 +588,7 @@ Recaptcha.configure do |config|
   config.secret_key = '6Lc6BAAAAAAAAKN3DRm6VA_xxxxxxxxxxxxxxxxx'
   config.verify_url = 'https://hcaptcha.com/siteverify'
   config.api_server_url = 'https://hcaptcha.com/1/api.js'
-  config.response_limit = 100000
```
grosser, line 591: that would break anyone already using that field, how about adding `response_minimum = 100` instead ?

test/verify_enterprise_test.rb
```diff
@@ -180,7 +180,7 @@ def initialize
       assert_equal "reCAPTCHA verification failed, please try again.", @controller.flash[:recaptcha_error]
     end
 
-    it "does not verify via http call when response length exceeds G_RESPONSE_LIMIT" do
+    it "does not verify via http call when response length exceeds G_RESPONSE_MAX_LIMIT" do
```
grosser, line 183: don't think this should change, the name references the var we send to recapcha ... possibly should just be rewritten to "exceeds limit"

grosser (approved):

PR #466, opened 2025-07-18
Add Reply class to handle Recaptcha responses

RubenIgnacio:
Add `Recaptcha::Reply` class to standardize and simplify response handling for both free and enterprise versions. It also defines the `[]=` method for backward compatibility.

### DIFF
--- a/lib/recaptcha.rb
+++ b/lib/recaptcha.rb
@@ -6,6 +6,7 @@ require 'uri'
 
 require 'recaptcha/configuration'
 require 'recaptcha/helpers'
+require 'recaptcha/reply'
 require 'recaptcha/adapters/controller_methods'
 require 'recaptcha/adapters/view_methods'
 if defined?(Rails)
@@ -76,15 +77,14 @@ module Recaptcha
     body['event']['expectedAction'] = options[:action] if options.key?(:action)
     body['event']['userIpAddress'] = options[:remote_ip] if options.key?(:remote_ip)
 
-    reply = api_verification_enterprise(query_params, body, project_id, timeout: options[:timeout])
-    score = reply.dig('riskAnalysis', 'score')
-    token_properties = reply['tokenProperties']
-    success = !token_properties.nil? &&
-      token_properties['valid'].to_s == 'true' &&
-      hostname_valid?(token_properties['hostname'], options[:hostname]) &&
-      action_valid?(token_properties['action'], options[:action]) &&
-      score_above_threshold?(score, options[:minimum_score]) &&
-      score_below_threshold?(score, options[:maximum_score])
+    raw_reply = api_verification_enterprise(query_params, body, project_id, timeout: options[:timeout])
+    reply = Reply.new(raw_reply, enterprise: true)
+
+    success = reply.success? &&
+      hostname_valid?(reply.hostname, options[:hostname]) &&
+      action_valid?(reply.action, options[:action]) &&
+      score_above_threshold?(reply.score, options[:minimum_score]) &&
+      score_below_threshold?(reply.score, options[:maximum_score])
 
     if options[:with_reply] == true
       [success, reply]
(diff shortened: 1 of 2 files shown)


lib/recaptcha/reply.rb
```diff
@@ -0,0 +1,77 @@
...
+    def initialize(raw_reply, enterprise: false)
+      @raw_reply = raw_reply
+      @enterprise = enterprise
+    end
+
+    def [](key)
+      return send(key.to_sym) if respond_to?(key.to_sym)
+
+      raw_reply[key]
+    end
```
grosser, line 14: why do we need this if the reply is not returned to the user ?

lib/recaptcha/reply.rb
```diff
@@ -0,0 +1,77 @@
...
+    def success?
+      success = enterprise? ? token_properties&.dig('valid') : raw_reply['success']
+
+      success.to_s == 'true'
+    end
+
+    def hostname
+      return raw_reply['hostname'] unless enterprise?
+
+      token_properties&.dig('hostname')
```
grosser, line 29: might be cleaner as `if enterprise? x else y end`

lib/recaptcha.rb
```diff
@@ -76,15 +77,14 @@ def self.verify_via_api_call_enterprise(response, options)
...
-      score_above_threshold?(score, options[:minimum_score]) &&
-      score_below_threshold?(score, options[:maximum_score])
+    raw_reply = api_verification_enterprise(query_params, body, project_id, timeout: options[:timeout])
+    reply = Reply.new(raw_reply, enterprise: true)
+
+    success = reply.success? &&
+      hostname_valid?(reply.hostname, options[:hostname]) &&
+      action_valid?(reply.action, options[:action]) &&
+      score_above_threshold?(reply.score, options[:minimum_score]) &&
+      score_below_threshold?(reply.score, options[:maximum_score])
```
grosser, line 87: can be dried up since they are the same now ?

lib/recaptcha/reply.rb
RubenIgnacio: I think there might be cases like mine, I access to the reply keys for debugging. It’s mainly during the transition, but I can add a warning or if it’s not needed, I can remove it.

lib/recaptcha.rb
RubenIgnacio: yes, I can put that validations in the class

grosser: was more thinking as a reused method in recaptcha.rb but in the class would also work

lib/recaptcha/reply.rb
grosser: would prefer removing, during debugging people can us the .to_h, but no big deal

lib/recaptcha/reply.rb
```diff
@@ -0,0 +1,77 @@
...
+      token_properties&.dig('createTime')
+    end
+
+    def enterprise?
+      @enterprise
+    end
+
+    def free?
+      !enterprise?
+    end
```
grosser, line 63: would actually avoid having 2 different methods, so the code is easier to reason about since it has less concepts to keep in mind

lib/recaptcha.rb
RubenIgnacio: i moved the validation to the `success?` method, then I added the `success` method to allow access to the raw value of that key. What do you think?

lib/recaptcha/reply.rb
RubenIgnacio: so remove the `free?` method?

lib/recaptcha/reply.rb
```diff
@@ -0,0 +1,116 @@
+module Recaptcha
+  class Reply
+    def initialize(raw_reply, enterprise: false)
```
grosser, line 3:
```suggestion
    def initialize(raw_reply, enterprise:)
```

lib/recaptcha/reply.rb
```diff
@@ -0,0 +1,116 @@
...
+      @raw_reply = raw_reply
+      @enterprise = enterprise
+    end
+
+    def token_properties
+      @raw_reply['tokenProperties'] if enterprise?
+    end
+
+    def success?(options = {})
```
grosser, line 12: having a `?` return a non-bool is a bit sus
```suggestion
    def success(options = {})
```

lib/recaptcha/reply.rb
grosser: I would remove it to simplify

grosser (approved):

grosser:
just some rubocop, rest seems fine 🎉

lib/recaptcha/reply.rb
RubenIgnacio: right, in that case we can keep `success?` like a boolean for test `success` and add `validate_success` method. What do you think?

RubenIgnacio:
> just some rubocop, rest seems fine 🎉

it seems it's because I forgot to add `# frozen_string_literal: true`. I'll add it now

grosser:
I'll do the polish :)
... do you want this released or this was just for fun refactor ?

RubenIgnacio:
A bit of both, initially I was just going to report the error message
problem with the recaptcha enterprise and in the process came up with a
solution

grosser:
5.20.0

PR #472, opened 2025-09-17
Allow flash messages when responding to turbo_stream requests

jrodriigues:
Rails form submissions and redirects are processed as TURBO_STREAM format and not HTML, this means that when calling `verify_recaptcha` without passing a model doesn't create a flash message.

## Pre-Merge Checklist
- [&#x2611;] CHANGELOG.md updated with short summary for user facing changes


grosser (approved):
nice!

PR #473, opened 2025-09-18
Fix changelog tag

jrodriigues:
@grosser thank you for getting #472 merged!

However I noticed that the changelog entry doesn't match the tag, so just dropping a quick fix to get that sorted.

## Pre-Merge Checklist
- [&#x2611;] CHANGELOG.md updated with short summary for user facing changes


grosser:
thx I'll add 5.21 too

